PR #1175, opened 2015-09-16
Added Ability to Import Resources from LDAP (Simple) Currently using

bghost4:
active Directory as a Test Case

Created Dialog to Import from LDAP with Server, BaseDN, Filter,
Username, and Password

Current Support, Imports Resources with Display Name Field, Omits null
results, does not set, Roles

TODO
	Import Mail, and Phone Numbers if Available from LDAP.
	Bring up a List of people to import in a JTable with a Checkbox Column,
ask Developers best/Standard way to do this. 
	Support Importing Active Directory/Ldap fields into Custom Fields.
Maybe Add some kind of Mapping to facilitate this.

### DIFF
--- a/ganttproject/data/resources/language/i18n.properties
+++ b/ganttproject/data/resources/language/i18n.properties
@@ -504,6 +504,7 @@ resource.copy.prefix = {0}_{1}
 resource.delete = $Delete Resource...
 resource.delete.description = Delete resource
 resource.import = Import Resources...
+resource.importLDAP = Import from Ldap
 resource.move.down = Down
 resource.move.down.description = Move resource down
 resource.move.up = Up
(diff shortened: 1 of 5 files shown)


bghost4:
I couldn't find anything on how to make this a plugin, which is probably the direction I would want to go with this. This is pretty rough around the edges at the moment, I was hoping that this pull request could at least begin some conversation of adding LDAP/AD import for people in domain environments to import and employee list from the Directory.

dbarashev:
Hey,
thanks for the patch.

May I ask you to format the code using 2 spaces instead of tabs? If you use Eclipse, you can use our formatter settings: https://github.com/bardsoftware/ganttproject/blob/master/ganttproject-builder/formatter.xml

Also, please add GPLv3 license notice to the beginning of the new files and add your name in @author tag in class javadoc comments.

Any instructions how to test this on Linux?

bghost4:
I work on Linux anyway, and this should be platform independent as far as I know, You would need access to an Active Directory server (Domain) server for this to work. Ideally a Generic LDAP server should work as well. I am working on cleaning up some rough edges and adding an import list to select just the people you want from active directory so It doesn't dump all users. I will update the formatting per your request and Add GPL headers. Coming in my next commit.

PR #1226, opened 2016-03-23
Updated GPLv2 and LGPLv2.1 license files

fsundermeyer:
The FSF address has changed. Replaced both files with versions
downloaded from the FSF.

I am about to create ganttproject packages for openSUSE and stumbled upon warnings from our build system about outdated versions of the GPL 2 and LGPL 2.1 (both contain an FSF address that is no longer valid).

Affected are

* GPLv2: org.ganttproject.impex.htmlpdf/resource/html-export-themes/striped_blue/COPYING
* LGPLv2.1: ganttproject/lib/core/licenses/jdnc.LICENSE


dbarashev:
Thanks!

Issue #1237, opened 2016-04-22
Disable "really delete" confirmation


ByronWillis:
When deleting a task you get a popup dialogue to confirm you "really want to delete".  With the ability to simply undo if you made a mistake, I think that this is an unnecessary question.

PR #1280, opened 2016-07-06
Create/Delete resource usability

dbarashev:
* Do not ask for confirmation when deleting
* Focus name when creating


dbarashev:
Update issue #1237

PR #1364, opened 2017-03-13
Use '-a' operator as 'AND' instead of '&&' in deb/postinst script

hiraku:
On installing ganttproject_2.8.3-r2088-1_all.deb, I found an error bellow is shown.
```
/var/lib/dpkg/info/ganttproject.postinst: 11: [: missing ]
```
In `ganttproject.postinst`, an operaotr `&&` is used in `[` command, but it should be `-a`.


dbarashev:
Hi, thanks for reporting this issue! It seems that `-a` operator is [deprecated](http://pubs.opengroup.org/onlinepubs/9699919799//utilities/test.html), though. The right way of fixing is to close left expression with `]` and open right expression with `[` (done in my commit above).

hiraku:
Thank you very much for your comment and more appropriate fix.
I did not know that `-a` is deprecated, thanks!

PR #1368, opened 2017-03-15
Prj2 excel start

MrBlackwell:


ganttproject/src/net/sourceforge/ganttproject/export/ExporterToExcel.java
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Created by valentin on 15.03.17.
+ */
+public class ExporterToExcel extends ExporterBase {
+
+  private static String[] FILE_EXTENTION = new String[] { ".xlsx" };
+
+  @Override
+  public String getFileTypeDescription() {
+    return GanttLanguage.getInstance().getText("");//add description in prorerty files and paste key here
```
dbarashev: If you meant to add todo to yourself, please prefix the comment with `TODO:`
Also, add a space between `;` and `//`.
Typo: s/prorerty/property

ganttproject/src/net/sourceforge/ganttproject/export/ExporterToExcel.java
```diff
@@ -0,0 +1,54 @@
+package net.sourceforge.ganttproject.export;
+
+import biz.ganttproject.core.option.GPOptionGroup;
+import net.sourceforge.ganttproject.language.GanttLanguage;
+
+import java.awt.*;
```
dbarashev: You need just a single `Component` from the whole `awt`, it makes sense to remove wildcard import

ganttproject/src/net/sourceforge/ganttproject/export/ExporterToExcel.java
```diff
@@ -0,0 +1,54 @@
...
+import java.io.File;
+import java.util.List;
+
+/**
+ * Created by valentin on 15.03.17.
```
dbarashev: We usually write `@author <email> (<your real name>)` or vice versa, `@author <Your real name> (<email>)`

dbarashev:
While the request itself looks good (see a few comments), are you sure that you really want to go this way? We discussed that Excel should be a file type option in _Spreadsheet_ export, similar to how image export looks. For that purpose you'll want to modify existing `ExporterToCSV` rather than add a new one.

Issue #907, opened 2015-06-28
Duration corresponding to percent complete in percent complete tooltip


GoogleCodeExporter:
```
Add new text to the progress bar tooltip.

When the user marks task's progress by dragging the black progress bar, there 
is a tooltip showing the progress percentage. Add new text to the same tooltip 
showing work done in days.
Example tooltip: "29%  5.2d"

Background:

When you work on a task, you most likely want to report amount of work done in 
days or hours rather than in percentages.

The current solution is: If you have a 15 day task and you used 2 days  to 
implement the task you first need to convert 2 days to percentages, which is 
13.3% and after that you can drag the black progress bar until progress has 
increased 13.3%.

The proposed solution is: If you have a 15 day task and you used 2 days  to 
implement the task you start dragging the progress bar and notice the initial 
amount in tooltip, e.g. 5d and drag until the amount in tooltip is 5+2=7days.

Remember that if you have zoomed out enough the Gantt chart, it shows weeks on 
the x-axis, not days and especially in this case additional text in tooltip is 
very helpful.

```

Original issue reported on code.google.com by `jk3m...@gmail.com` on 18 Mar 2014 at 1:53

Issue #225, opened 2015-06-28
Duration in hours


GoogleCodeExporter:
```
What steps will reproduce the problem?
1. When adding a task, it is assumed that the duration is one.  It isn't 
possible to reduce this below 1.  The measurement is in days - what if there is 
a half day or 3 hour task?

It would be good to be able to indicate 1hr, 3hrs or 7hrs or 1 day.  I guess 
Hours, Days, Months would be the desired units.


What version of the product are you using? On what operating system?
Using 2.0.10 on Windows 7.

```

Original issue reported on code.google.com by `Harborou...@gmail.com` on 19 Apr 2011 at 2:38

RadialAdmin:
I update the issue, to recall that it is really important to many users.

ravelus:
Cannot believe this is not implemented yet after *years* of divagation... this is a must!!!

griznold:
Is there any update? I would really appreciate this feature, too.

GRA0007:
Yes it would be good to have this feature. At the moment I have to use excel for projects that use minutes or hours. Does anyone have an update on the progress of this issue?

Hillside502:
The duration less than a day (MISSING) "feature" is a total and utter dealbreaker.

It has been debated/requested/IGNORED for over 12 years, as follows:-

GanttProject / Feature Requests / #34 Hour/Minutes/Seconds granularity of task duration
https://sourceforge.net/p/ganttproject/feature-requests/34/

Currently, this follows on at:-

Duration in hours · Issue #225 · bardsoftware/ganttproject
https://github.com/bardsoftware/ganttproject/issues/225

Why on earth anyone/everyone would NOT want tasks with a duration of LESS THAN A DAY is beyond me.

Bye, Bye GanttProject --- dead in the water --- before I give it any more of my "time" and "effort"!

sashimiFR:
I'm discovering Ganttproject, and immediately ran into this duration granularity issue too. Be it dealing with professional projects in my consulting work, or trying to optimise how I'm going to do the 3 layers of painting of each 2 sides of the 12 shutters of my home, adequate timescaling is brick one of any project management with a chance of success... It can't and shouldn't be a random "feature" which can be delayed for over 10 years. And being free software is no excuse. If you're going to build a project management app, do it properly or don't bother doing it at all (I mean, really, what's the point ?).

[...]

PR #1371, opened 2017-03-22
Fix for #907 (Duration corresponding to percent complete in percent complete tooltip)

okusha:
This is a quick fix for #907. It lacks i18n (for 'd' symbol or the whole hint message pattern) but my view is that this should be done on Task level i.e. Task.getProgressString() localized representation of progress in task units. Given potential future implementation of #225 I hesitated to make changes in task package just to get hint's text changed.


dbarashev:
Hi Oleg,
what do you think of encapsulating this in `ChangeTaskProgressRuler` class which is used for calculating percentage shown in the hint? We could return a data object instead of `int` from there and could even abstract from duration units -- at least use `TimeUnitStack.encode` method to get that `d` in the label :)

By the way, I suggest writing `3+ d` instead of `3.14 d`. It is not clear what "0.14 d" means anyway. Or maybe '3d', '3 1/2 d', '4d', that is, increment duration at half-day intervals

okusha:
Hi Dmitry,

- task progress conversion is moved to `ChangeTaskProgressRuler`,
 data object `ChangeTaskProgressRuler.Progress` is introduced, 
- task unit symbol is taken from `TimeUnitStack.encode`, 
- test `ChangeTaskProgressRulerTest.testProgressConversionToUnits` is introduced

I chose `3+d` pattern for now.

ganttproject/src/net/sourceforge/ganttproject/chart/mouse/ChangeTaskProgressRuler.java
```diff
@@ -88,6 +88,34 @@ int getProgress(int pixels) {
...
+
+  String convertPercentsToUnits(int overallProgress) {
+    float taskDuration = myTask.getDuration().getValue();
+    float progressInUnits = overallProgress * taskDuration/100;
+    String wholeUnits = Integer.toString((int)progressInUnits);
+    String fractionIndicator = (overallProgress*myTask.getDuration().getLength()) % 100 == 0 ? "" : "+";
+    return wholeUnits + fractionIndicator;
+  }
+
+  interface Progress {
```
dbarashev: Just class maybe? There is just one implementation anyway.

ganttproject/src/net/sourceforge/ganttproject/chart/mouse/ChangeTaskProgressRuler.java
```diff
@@ -88,6 +88,34 @@ int getProgress(int pixels) {
...
+        return convertPercentsToUnits(progressPercents);
+      }
+    };
+  }
+
+  String convertPercentsToUnits(int overallProgress) {
+    float taskDuration = myTask.getDuration().getValue();
+    float progressInUnits = overallProgress * taskDuration/100;
+    String wholeUnits = Integer.toString((int)progressInUnits);
+    String fractionIndicator = (overallProgress*myTask.getDuration().getLength()) % 100 == 0 ? "" : "+";
```
dbarashev: nit: surround `*` with spaces pls

ganttproject/src/net/sourceforge/ganttproject/chart/mouse/ChangeTaskProgressInteraction.java
```diff
@@ -61,15 +61,22 @@ private Task getTask() {
...
+    int newProgress = progress.toPercents();
     if (newProgress > 100) {
       newProgress = 100;
     }
     if (newProgress < 0) {
       newProgress = 0;
     }
     myMutator.setCompletionPercentage(newProgress);
-    myLastNotes = new TaskInteractionHintRenderer(newProgress + "%", event.getX(), event.getY() - 30);
+    String hintText =
```
dbarashev: You may like `MessageFormat` class or `String.format` method, and it probably makes sense to i18n it because every culture has its own understanding of proper spacing between numbers and percent signs (no kidding). I normally create a new key with formatting pattern like `{0} %({1}{2})` in `ganttproject/data/resources/language/i18n.properties` file. `GanttLanguage` is a handy singleton for accessing i18n keys and passing arguments.

ganttproject-tester/test/net/sourceforge/ganttproject/chart/mouse/ChangeTaskProgressRulerTest.java
```diff
@@ -49,6 +49,32 @@ private static Task createTask(TaskManager taskManager) {
         return result;
     }
 
+    public void testProgressConversionToUnits() {
+        TaskChartModelFacade mockChartModel = EasyMock.createMock(TaskChartModelFacade.class);
```
dbarashev: extract this code which creates a task, rectangles and ruler into a helper method? It is duplicated in two tests.

ganttproject/src/net/sourceforge/ganttproject/chart/mouse/ChangeTaskProgressRuler.java
okusha: Right

ganttproject/src/net/sourceforge/ganttproject/chart/mouse/ChangeTaskProgressRuler.java
okusha: I meant interface here as a less powerful version of class with compact syntax not as an API thing, maybe it was misleading. Converted it to abstract class now.

ganttproject/src/net/sourceforge/ganttproject/chart/mouse/ChangeTaskProgressInteraction.java
okusha: Done.

ganttproject/src/net/sourceforge/ganttproject/chart/mouse/ChangeTaskProgressRuler.java
okusha: Progress is a static class with all the conversion inside

ganttproject-tester/test/net/sourceforge/ganttproject/chart/mouse/ChangeTaskProgressRulerTest.java
okusha: `testProgressConversionToUnits` does not share text fixture with the other tests anymore as Progress class is tested directly without mock objects

ganttproject/data/resources/language/i18n.properties
```diff
@@ -616,6 +616,7 @@ task.move.up = Up
 task.move.up.description = Move task up
 task.new = New task
 task.notesTooltip.pattern = <html>{0}</html>
+task.changeProgressHint = {0}% ({1}d)
```
okusha: 1.   `TimeUnitStack.encode` was a good alternative to another hardcoded "d" symbol
2.  Currently `TimeUnitStack.encode` is not i18-ed and it's not clear to me if it should  as some parts of code could potentially rely on that hardcoded "d" in logic. If it's not to be i18n-ed than we would have "d" symbol in hint for all the languages
3.   If  `TimeUnitStack.encode`needs i18n for its strings then we can do that and pass localized "d" as {2} in the discussed string. Or we can leave it as is considering hint text as independently localized string.

dbarashev: Just pass here whatever `encode` returns :)

ganttproject/src/net/sourceforge/ganttproject/chart/mouse/ChangeTaskProgressRuler.java
```diff
@@ -88,6 +88,28 @@ int getProgress(int pixels) {
     int upperProgress = tailMap.get(upperPx);
 
     float diffProgress = (upperProgress - lowerProgress) * ((float) (pixels - lowerPx) / (float) (upperPx - lowerPx));
-    return (int) (lowerProgress + diffProgress);
+    int overallProgress = (int) (lowerProgress + diffProgress);
+    return new Progress(overallProgress, myTask.getDuration().getLength());
+  }
+
+  static class Progress {
+    private int myPercents;
```
dbarashev: make them final?

okusha: good catch!

ganttproject/src/net/sourceforge/ganttproject/chart/mouse/ChangeTaskProgressRuler.java
```diff
@@ -89,34 +89,27 @@ Progress getProgress(int pixels) {
...
+    private int myTaskDuraton;
 
-  String convertPercentsToUnits(int overallProgress) {
-    float taskDuration = myTask.getDuration().getValue();
-    float progressInUnits = overallProgress * taskDuration / 100;
-    String wholeUnits = Integer.toString((int) progressInUnits);
-    String fractionIndicator = (overallProgress * myTask.getDuration().getLength()) % 100 == 0 ? "" : "+";
-    return wholeUnits + fractionIndicator;
-  }
+    Progress(int percents, int taskDuration) {
```
okusha: agree

ganttproject/data/resources/language/i18n.properties
okusha: No problems, done. But will you explain why it's the best way?

dbarashev: should TaskDuration be in hours, it would insert `h` automatically. Just gettign rid of code, configs or whatever else which think that duration is measured in days.

PR #1375, opened 2017-03-24
Prj1 Getting calendars and created plugin

shatuln:
- Created new plugin and moved code from main package

- Getting writable calendars from user


biz.ganttproject.impex.google/plugin.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<plugin
+   id="biz.ganttproject.impex.google"
+   name="Google prj"
+   version="2.8.4"
+   provider-name="">
+
+   <runtime>
+      <library name="lib/core/google-http-client-1.22.0.jar">
```
dbarashev: Did you try to run this code? You have no `core` directory in this plugin. You have no `biz.ganttproject.impex.google.jar` in the runtime section either.

Issue #1373, opened 2017-03-22
Make it easy to assign/remove resource to/from many tasks


dbarashev:
Add "Assignments" tab to Resource Properties dialog

PR #1378, opened 2017-03-29
#1373 (Make it easy to assign/remove resource to/from many tasks)

okusha:
I'm not sure what should be a correct header for new files (copyrights, etc)

### DIFF
--- a/ganttproject/src/net/sourceforge/ganttproject/gui/ResourceAssignmentsPanel.java
+++ b/ganttproject/src/net/sourceforge/ganttproject/gui/ResourceAssignmentsPanel.java
@@ -1,0 +1,97 @@
+/*
+GanttProject is an opensource project management tool. License: GPL3
+Copyright (C) 2010 Dmitry Barashev
+
+This program is free software; you can redistribute it and/or
+modify it under the terms of the GNU General Public License
+as published by the Free Software Foundation; either version 3
+of the License, or (at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package net.sourceforge.ganttproject.gui;
+
+import net.sourceforge.ganttproject.resource.HumanResource;
+import net.sourceforge.ganttproject.task.ResourceAssignment;
+import net.sourceforge.ganttproject.task.Task;
+import net.sourceforge.ganttproject.task.TaskManager;
+
+import javax.swing.*;
+import javax.swing.table.TableColumn;
+
+/**
+ * UI component in a resource properties dialog: a table with tasks assigned to
+ * a resource.
+ *
+ * @author Oleg Kushnikov
+ */
+public class ResourceAssignmentsPanel {
+  private final TaskManager myTaskManager;
+  private ResourceAssignmentsTableModel myModel;
+  private final HumanResource myPerson;
+  private JTable myTable;
+
+  public ResourceAssignmentsPanel(HumanResource person, TaskManager taskManager) {
+    myPerson = person;
+    myTaskManager = taskManager;
+  }
+
+  private JTable getTable() {
+    return myTable;
+  }
+
+  public JPanel getComponent() {
+    myModel = new ResourceAssignmentsTableModel(myPerson);
+    myTable = new JTable(myModel);
+    UIUtil.setupTableUI(getTable());
+    setUpTasksComboColumn(getTable().getColumnModel().getColumn(ResourceAssignmentsTableModel.Column.NAME.ordinal()), getTable());
+    AbstractTableAndActionsComponent<ResourceAssignment> tableAndActions =
+        new AbstractTableAndActionsComponent<ResourceAssignment>(getTable()) {
+      @Override
+      protected void onAddEvent() {
+        getTable().editCellAt(myModel.getRowCount() - 1, 1);
+      }
+
+      @Override
+      protected void onDeleteEvent() {
+        if (getTable().isEditing() && getTable().getCellEditor() != null) {
+          getTable().getCellEditor().stopCellEditing();
+        }
+        myModel.delete(getTable().getSelectedRows());
+      }
+
+      @Override
+      protected ResourceAssignment getValue(int row) {
+        java.util.List<ResourceAssignment> values = myModel.getResourcesAssignments();
+        return (row >= 0 && row < values.size()) ? values.get(row) : null;
+      }
+    };
+    JPanel tablePanel = AbstractTableAndActionsComponent.createDefaultTableAndActions(getTable(), tableAndActions.getActionsComponent());
+    return tablePanel;
+  }
+
+  public void commit() {
+    if (myTable.isEditing()) {
+      myTable.getCellEditor().stopCellEditing();
+    }
+    myModel.commit();
+  }
+
+  protected void setUpTasksComboColumn(TableColumn column, final JTable table) {
+    final JComboBox comboBox = new JComboBox();
+    Task[] tasks = myTaskManager.getTasks();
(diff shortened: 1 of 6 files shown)


dbarashev:
File header:
```
/*
Copyright 2017 Oleg Kushnikov, BarD Software s.r.o

This file is part of GanttProject, an opensource project management tool.

GanttProject is free software: you can redistribute it and/or modify
it under the terms of the GNU General Public License as published by
 the Free Software Foundation, either version 3 of the License, or
 (at your option) any later version.

GanttProject is distributed in the hope that it will be useful,
but WITHOUT ANY WARRANTY; without even the implied warranty of
MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
GNU General Public License for more details.

You should have received a copy of the GNU General Public License
along with GanttProject.  If not, see <http://www.gnu.org/licenses/>.
*/
```

This assumes that both you and BarD Software are copyright holders with the same rights (including the right to re-license this code)

ganttproject/src/net/sourceforge/ganttproject/gui/ResourceAssignmentsTableModel.java
```diff
@@ -0,0 +1,179 @@
...
+import net.sourceforge.ganttproject.task.ResourceAssignmentMutator;
+import net.sourceforge.ganttproject.task.Task;
+
+import javax.swing.table.AbstractTableModel;
+import java.util.*;
+
+/**
+ * @author Oleg Kushnikov
+ */
+public class ResourceAssignmentsTableModel extends AbstractTableModel {
```
dbarashev: There is a few good comments from IDEA about visibility restrictions and simplifying code (yellow ticks in the editor scrollbar)

ganttproject/src/net/sourceforge/ganttproject/gui/ResourceAssignmentsPanel.java
```diff
@@ -0,0 +1,97 @@
...
+  private JTable getTable() {
+    return myTable;
+  }
+
+  public JPanel getComponent() {
+    myModel = new ResourceAssignmentsTableModel(myPerson);
+    myTable = new JTable(myModel);
+    UIUtil.setupTableUI(getTable());
+    setUpTasksComboColumn(getTable().getColumnModel().getColumn(ResourceAssignmentsTableModel.Column.NAME.ordinal()), getTable());
+    AbstractTableAndActionsComponent<ResourceAssignment> tableAndActions =
```
dbarashev: IDEA says this code is identical to code in TaskAllocationsPanel. It seems to be pretty abstract and only needs `delete` and `getAllValues()` which are not in the standard `TableModel`. Maybe just create `TableModelExt` and move this code into method `AbstractTableAndActionsComponent.create(JTable, TableModelExt)` ?

ganttproject/src/net/sourceforge/ganttproject/gui/ResourceAssignmentsTableModel.java
```diff
@@ -0,0 +1,179 @@
...
+
+  @Override
+  public int getRowCount() {
+    return myAssignments.size() + 1;
+  }
+
+  @Override
+  public boolean isCellEditable(int row, int col) {
+    if (row == myAssignments.size()) {
+      return true;
```
okusha: in the task description you shared with me originally this field was required to be editable. But I agree that if we consider the primary goal o this page to be high-level mapping of the resource to tasks than add/delete  should be enough of edit possibilities

ganttproject/src/net/sourceforge/ganttproject/gui/ResourceAssignmentsPanel.java
```diff
@@ -0,0 +1,97 @@
...
+      protected void onDeleteEvent() {
+        if (getTable().isEditing() && getTable().getCellEditor() != null) {
+          getTable().getCellEditor().stopCellEditing();
+        }
+        myModel.delete(getTable().getSelectedRows());
+      }
+
+      @Override
+      protected ResourceAssignment getValue(int row) {
+        java.util.List<ResourceAssignment> values = myModel.getResourcesAssignments();
```
okusha: Done. Actually, Idea has the right setting on switched on by default (Settings -> Editor -> Intentions -> Java -> Imports -> Replace On Demand import Single Class imports). The case here was that I used cut-n-paste technology to start with the class and the original code contained java.swing.* import which was not converted by Idea for some reason :)

ganttproject/src/net/sourceforge/ganttproject/gui/ResourceAssignmentsPanel.java
```diff
@@ -0,0 +1,97 @@
...
+        return (row >= 0 && row < values.size()) ? values.get(row) : null;
+      }
+    };
+    JPanel tablePanel = AbstractTableAndActionsComponent.createDefaultTableAndActions(getTable(), tableAndActions.getActionsComponent());
```
okusha: done

ganttproject/src/net/sourceforge/ganttproject/gui/ResourceAssignmentsPanel.java
```diff
@@ -0,0 +1,97 @@
...
+    if (myTable.isEditing()) {
+      myTable.getCellEditor().stopCellEditing();
+    }
+    myModel.commit();
+  }
+
+  protected void setUpTasksComboColumn(TableColumn column, final JTable table) {
+    final JComboBox comboBox = new JComboBox();
+    Task[] tasks = myTaskManager.getTasks();
+    for (int i = 0; i < tasks.length; i++) {
```
okusha: done

ganttproject/src/net/sourceforge/ganttproject/gui/ResourceAssignmentsTableModel.java
okusha: agreed with Idea in most of the cases

ganttproject/src/net/sourceforge/ganttproject/gui/ResourceAssignmentsTableModel.java
okusha: Don, only name of new task can be edited now

ganttproject/src/net/sourceforge/ganttproject/gui/ResourceAssignmentsTableModel.java
```diff
@@ -0,0 +1,179 @@
...
+
+  @Override
+  public int getColumnCount() {
+
+    return Column.values().length;
+  }
+
+  @Override
+  public Object getValueAt(int row, int col) {
+    assert row >= 0 && row < getRowCount() && col >= 0 && col < getColumnCount();
```
okusha: done

ganttproject/src/net/sourceforge/ganttproject/gui/ResourceAssignmentsTableModel.java
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  @Override
+  public int getRowCount() {
+    return myAssignments.size() + 1;
+  }
+
+  @Override
+  public boolean isCellEditable(int row, int col) {
+    return row == myAssignments.size() && Column.NAME.equals(Column.values()[col]);
```
dbarashev: No, I didn't mean that the only thing which is editable is task name in the new row :)
I meant that in the new row the only thing which is editable is task name. In the other rows load should be editable (and task name probably should not, but I am not really sure)

okusha: Right, my bad. 
Only task name is editable for new row and only load is editable for "old" rows for now.

ganttproject/src/net/sourceforge/ganttproject/gui/ResourceAssignmentsTableModel.java
```diff
@@ -0,0 +1,179 @@
...
+import java.util.*;
+
+/**
+ * @author Oleg Kushnikov
+ */
+public class ResourceAssignmentsTableModel extends AbstractTableModel {
+  enum Column {
+    ID("id", String.class),
+    NAME("taskname", String.class),
+    UNIT("unit", String.class);
```
okusha: It's Float now with the same validating capabilities. Chose Float as float is already used as data type for `setLoad()` and `getLoad()` in `ResourceAssignment`

PR #1381, opened 2017-03-30
Implemented sorting

Kambius:
TODO: add arrows for indicating order


ganttproject/src/net/sourceforge/ganttproject/TaskContainmentHierarchyFacadeImpl.java
```diff
@@ -136,6 +124,52 @@ public Task getContainer(Task nestedTask) {
...
 
   @Override
+  public void sort(Comparator<Task> comparator) {
+    sortHelper(getRootTask(), comparator);
+  }
+
+  private void sortHelper(Task root, Comparator<Task> comparator) {
+    Task[] tasks = getNestedTasks(root);
+    // Bubble sort
+    for (int i = 0; i < tasks.length; i++) {
```
dbarashev: If I were you I would just remove all nodes from parent, sort them using `Collections.sort` and add back to the parent in the sorted order. This will make the code more simple and less fragile.

ganttproject/src/net/sourceforge/ganttproject/GPTreeTableBase.java
```diff
@@ -528,10 +536,80 @@ protected void initTreeTable() {
     doInit();
   }
 
+  private static Comparator<Task> beginComparatorAsc = new Comparator<Task>() {
```
dbarashev: Please move comparators into the table column class. It was mentioned in the [design document](https://github.com/bardsoftware/ganttproject/wiki/Feature-Notes:-Task-Sorting):
> add sortComparator property to TaskDefaultColumn interface and its implementation in BEGIN_DATE and END_DATE

and it will encapsulate column logic in a single place.

ganttproject/src/net/sourceforge/ganttproject/GPTreeTableBase.java
```diff
@@ -528,10 +536,80 @@ protected void initTreeTable() {
...
+    public int compare(Task t1, Task t2) {
+      return -t1.getEnd().compareTo(t2.getEnd());
+    }
+  };
+  
   protected void doInit() {
     setRootVisible(false);
     myCustomPropertyManager.addListener(this);
 
+    getTree().getTableHeader().addMouseListener(new MouseAdapter() {
```
dbarashev: What's the purpose of this listener?

ganttproject/src/net/sourceforge/ganttproject/GPTreeTableBase.java
```diff
@@ -528,10 +536,80 @@ protected void initTreeTable() {
...
+      @Override
+      public void mouseClicked(MouseEvent mouseEvent) {
+        int index = getTable().columnAtPoint(mouseEvent.getPoint());
+        ColumnImpl column = myTableHeaderFacade.findColumnByViewIndex(index);
+
+
+        if (column.myStub.getID().equals(TaskDefaultColumn.BEGIN_DATE.getStub().getID())) {
+          if (column.getSort() > 0) {
+            column.setSort(-1);
+            myProject.getTaskManager().getTaskHierarchy().sort(beginComparatorDesc);
```
dbarashev: and if comparator was in the TaskDefaultColumn you could just write here `column.myStub.getComparator(column.getSort())`

ganttproject/src/net/sourceforge/ganttproject/GPTreeTableBase.java
```diff
@@ -346,13 +345,22 @@ protected ColumnImpl findColumnByViewIndex(int index) {
...
+    private int sort;
 
     protected ColumnImpl(JXTreeTable table, TableColumnExt tableColumn, ColumnList.Column stub) {
       myTable = table;
       myTableColumn = tableColumn;
       myStub = stub;
     }
 
+    public int getSort() {
+      return sort;
```
dbarashev: please follow [our style guide](https://docs.ganttproject.biz/development/#_change_process): `mySort`. I also recommend using either boolean or (better) enum for sorting. `getSort() == SortOrder.ASC` is way more readable than `getSort() == 1`

ganttproject/src/net/sourceforge/ganttproject/GPTreeTableBase.java
Kambius: Sory, I forgot to delete it

ganttproject/src/net/sourceforge/ganttproject/task/TaskManagerImpl.java
```diff
@@ -45,49 +23,29 @@
...
 import net.sourceforge.ganttproject.task.dependency.constraint.FinishFinishConstraintImpl;
 import net.sourceforge.ganttproject.task.dependency.constraint.FinishStartConstraintImpl;
 import net.sourceforge.ganttproject.task.dependency.constraint.StartFinishConstraintImpl;
 import net.sourceforge.ganttproject.task.dependency.constraint.StartStartConstraintImpl;
-import net.sourceforge.ganttproject.task.event.TaskDependencyEvent;
-import net.sourceforge.ganttproject.task.event.TaskHierarchyEvent;
-import net.sourceforge.ganttproject.task.event.TaskListener;
-import net.sourceforge.ganttproject.task.event.TaskPropertyEvent;
-import net.sourceforge.ganttproject.task.event.TaskScheduleEvent;
+import net.sourceforge.ganttproject.task.event.*;
```
dbarashev: Please setup your editor so that it didn't replace imports with wildcard. If you're using IntelliJ IDEA, it is "Class count to use import with '*'" in Editor->Code Style->Java

These diffs just clutter the pull request.

ganttproject/src/net/sourceforge/ganttproject/GPTreeTableBase.java
Kambius: I don't understand how I can get TaskDefaultColumn from col index in GPTreeTableBase, could you help my with it&

dbarashev: if you can get the ID then you can call `TaskDefaultColumn.find(id)`

ganttproject/src/net/sourceforge/ganttproject/GPTreeTableBase.java
```diff
@@ -346,13 +345,22 @@ protected ColumnImpl findColumnByViewIndex(int index) {
     private final JXTreeTable myTable;
     private final TableColumnExt myTableColumn;
     private final Column myStub;
+    private SortOrder sort;
```
dbarashev: `mySort` pls

ganttproject/src/net/sourceforge/ganttproject/GPTreeTableBase.java
```diff
@@ -528,10 +536,97 @@ protected void initTreeTable() {
...
+
+  private Comparator<Task> endComparatorDesc = new Comparator<Task>() {
+    @Override
+    public int compare(Task t1, Task t2) {
+      return -t1.getEnd().compareTo(t2.getEnd());
+    }
+  };
+
+  private class SortTableHeaderRenderer implements TableCellRenderer {
+    private Icon ascIcon;
```
dbarashev: Please use `my` prefix: `myAscIcon`

Also, you may want to make these fields final.

ganttproject/src/net/sourceforge/ganttproject/GPTreeTableBase.java
```diff
@@ -528,10 +536,97 @@ protected void initTreeTable() {
...
+      ascIcon = UIManager.getIcon("Table.ascendingSortIcon");
+      descIcon = UIManager.getIcon("Table.descendingSortIcon");
+    }
+
+    @Override
+    public Component getTableCellRendererComponent(JTable table, Object value, boolean isSelected, boolean hasFocus, int row, int col) {
+      Component c = defaultRenderer.getTableCellRendererComponent(table, value, isSelected, hasFocus, row, col);
+      ColumnImpl column = myTableHeaderFacade.findColumnByViewIndex(col);
+      if (column.getSort() == SortOrder.ASCENDING) {
+        ((JLabel) c).setIcon(ascIcon);
```
dbarashev: Please add `instanceof JLabel` check. If default component is not JLabel then we just ignore it (or maybe replace with our own JLabel?)

ganttproject/src/net/sourceforge/ganttproject/GPTreeTableBase.java
```diff
@@ -565,24 +565,28 @@ public int compare(Task t1, Task t2) {
...
     }
 
     @Override
     public Component getTableCellRendererComponent(JTable table, Object value, boolean isSelected, boolean hasFocus, int row, int col) {
-      Component c = defaultRenderer.getTableCellRendererComponent(table, value, isSelected, hasFocus, row, col);
+      Component c = myDefaultRenderer.getTableCellRendererComponent(table, value, isSelected, hasFocus, row, col);
       ColumnImpl column = myTableHeaderFacade.findColumnByViewIndex(col);
       if (column.getSort() == SortOrder.ASCENDING) {
-        ((JLabel) c).setIcon(ascIcon);
+        if (c instanceof JLabel) {
```
dbarashev: Please move this check to upper level, before you start dealing with columns and sort orders.

ganttproject/src/net/sourceforge/ganttproject/GPTreeTableBase.java
Kambius: I have no access to Task in TaskDefaultColumn bacause it is in other module, so I can't create comparator there

dbarashev: Good point. You can use the same approach which is used with isEditable. Look at `GanttTreeTableModel` constructor.

dbarashev:
Bug report: when table is sorted by e.g. start date (and you see the arrow indicator) and you click "end date" column, you get two arrows.

dbarashev:
One more bugreport: arrow won't hide after adding/removing/moving a task (they should because we do not maintain the order)

ganttproject/src/net/sourceforge/ganttproject/GPTreeTableBase.java
```diff
@@ -528,10 +536,72 @@ protected void initTreeTable() {
...
+        int index = getTable().columnAtPoint(mouseEvent.getPoint());
+        if (index == -1) return;
+
+        ColumnImpl column = myTableHeaderFacade.findColumnByViewIndex(index);
+        TaskDefaultColumn taskColumn = TaskDefaultColumn.find(column.getID());
+
+        if (taskColumn == TaskDefaultColumn.BEGIN_DATE || taskColumn == TaskDefaultColumn.END_DATE) {
+          if (column.getSort() == SortOrder.ASCENDING) {
+            column.setSort(SortOrder.DESCENDING);
+            getTree().getColumn(index).setHeaderRenderer(new SortTableHeaderRenderer(getTable()));
```
dbarashev: is it required to change header renderer every time? Can it be set just once?

Kambius:
All tasks finished

biz.ganttproject.core/src/biz/ganttproject/core/model/task/TaskDefaultColumn.java
```diff
@@ -76,6 +78,14 @@ private TaskDefaultColumn(ColumnList.Column delegate, Class<?> valueClass, Strin
     myNameKey = nameKey;
   }
 
+  public Comparator<?> getSortComparator() {
+    return mySortComparator;
+  }
+
+  public void setSortComparator(Comparator<?> mySortComparator) {
+    this.mySortComparator = mySortComparator;
```
dbarashev: The purpose of prefixing with `my` is to indicate that it is a member variable without IDE assistance. Additional prefixing with `this.` is redundant: we already know it is a member variable.

ganttproject/src/net/sourceforge/ganttproject/GPTreeTableBase.java
```diff
@@ -528,10 +544,139 @@ protected void initTreeTable() {
...
+      myAscIcon = UIManager.getIcon("Table.ascendingSortIcon");
+      myDescIcon = UIManager.getIcon("Table.descendingSortIcon");
+    }
+
+    @Override
+    public Component getTableCellRendererComponent(JTable table, Object value, boolean isSelected, boolean hasFocus, int row, int col) {
+      Component c = myDefaultRenderer.getTableCellRendererComponent(table, value, isSelected, hasFocus, row, col);
+      ColumnImpl column = myTableHeaderFacade.findColumnByViewIndex(col);
+      if (column.getSort() == SortOrder.ASCENDING) {
+        if (c instanceof JLabel) {
```
dbarashev: Please move `instanceof` check upper. Do it prior to dealing with columns. This will make code cleaner, just by removing multiple nested ifs.

ganttproject/src/net/sourceforge/ganttproject/GPTreeTableBase.java
```diff
@@ -528,10 +544,139 @@ protected void initTreeTable() {
...
+    public void taskModelReset() {
+      removeOrder();
+    }
+  };
+
   protected void doInit() {
     setRootVisible(false);
     myCustomPropertyManager.addListener(this);
 
+    tableHeader.setDefaultRenderer(new SortTableHeaderRenderer(getTable()));
```
dbarashev: Normally you want to access protected members from the superclass by calling accessor function, like you do one line below

ganttproject/src/net/sourceforge/ganttproject/GPTreeTableBase.java
```diff
@@ -528,10 +544,139 @@ protected void initTreeTable() {
...
+
+    getTree().getTableHeader().addMouseListener(new MouseAdapter() {
```
dbarashev: `getTree()` is not necessary here, it return `this`.

ganttproject/src/net/sourceforge/ganttproject/GPTreeTableBase.java
```diff
@@ -528,10 +544,139 @@ protected void initTreeTable() {
...
+      if (column.getSort() == SortOrder.DESCENDING) {
+        if (c instanceof JLabel) {
+          ((JLabel) c).setIcon(myDescIcon);
+        }
+      }
+      return c;
+    }
+  }
+
+  private <T>Comparator<T> reverseComparator(final Comparator<T> comparator) {
```
dbarashev: please make it static

ganttproject/src/net/sourceforge/ganttproject/GPTreeTableBase.java
```diff
@@ -528,10 +544,139 @@ protected void initTreeTable() {
...
+    }
+
+    @Override
+    public void taskPropertiesChanged(TaskPropertyEvent e) {
+      removeOrder();
+    }
+
+    @Override
+    public void taskProgressChanged(TaskPropertyEvent e) {
+      removeOrder();
```
dbarashev: Properties and progress do not affect task dates and shall not reset the ordering. In fact, listening to `taskScheduleChanged, taskAdded, taskRemoved and taskModelReset` should be enough to reset ordering. You may want to use `TaskListenerAdapter` with default empty implementations of all methods.

ganttproject/src/net/sourceforge/ganttproject/GPTreeTableBase.java
```diff
@@ -346,13 +353,22 @@ protected ColumnImpl findColumnByViewIndex(int index) {
...
       myTableColumn = tableColumn;
       myStub = stub;
     }
 
+    public SortOrder getSort() {
+      return mySort;
+    }
+
+    public void setSort(SortOrder sort) {
+      this.mySort = sort;
```
dbarashev: Same about `this`

PR #1387, opened 2017-04-03
Feature export to xls

akurutin:
Небольшой рефакторинг для того. чтобы в дальнейшем можно было имплементить простые writer`ы(в том числе xls). Судя по всему, master и prj2_master - не очень совпадают. И build.gradle в винде дает ошибку - жалуется 

Where:
Build file 'C:\src\gantproject\ganttproject-builder\build.gradle' line: 65

What went wrong:
Execution failed for task ':..:ganttproject:updateMavenDeps'.

file could`t be normalized (скорее всего как раз из-за '*')


dbarashev:
Please use `prj2_master` as your base branch. To be clear: it means not only that you need to build pull request against this branch (which you did), but also that you should fork from this branch (which you didn't, and forked from `master` instead).

dbarashev:
`master` is now fixed and I merged it into `prj2_master`. Anyway, please use `prj2_master` as your base branch.

akurutin:
I almost did, just transferred the body to an abstract class and made 2 writers with print and println methods.Do you want to make export to xls a part of exporting to csv?

PR #1389, opened 2017-04-04
Add Kotlin to GanttProject build

dbarashev:


okusha:
Gradle cannot build it for me:
```
:..:ganttproject:compileKotlin
Using kotlin incremental compilation
w: The '-d' option with a directory destination is ignored because '-module' is specified
e: C:\src\gantproject\ganttproject\src\net\sourceforge\ganttproject\HelloKotlin.kt: (12, 5): Unresolved reference: println
:..:ganttproject:compileKotlin FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':..:ganttproject:compileKotlin'.
> Compilation error. See log for more details
```
It seems that Kotlin is not configured properly in `gradle.build`
I can make it working if I configure Kotlin for the module in Idea but this leads to a changed `gradle.build`

Issue #130, opened 2015-06-28
Show task completion speed somewhere


GoogleCodeExporter:
```
Hi.

Is it possible to assign formulaes to a new personalized column? If so, forget 
the feature request. But if not, here´s what I´d like:

I want to calculate how well am I doing about the schedule. If a task is 
estimated for 20 days, and today is day 10th, I should be at 50%, right? But if 
I am at 80% in progress, that means that I am 80% advanced. That 80% is the 
number I want Gantt to calculate me, or a way to do so..

That´s all, thanks and excuse my bad English.


```

Original issue reported on code.google.com by `nsieg...@gmail.com` on 23 Jun 2010 at 5:49

Issue #1309, opened 2016-10-04
Change logs directory


m0squito:
I'm using GP 2.8.1 (Pilsen) on macOS 10.12. Each time I start the app, it creates (if not already existing) two files: ganttproject.log and java0.log directly in my home directory. Is there a way to a) disable the log altogether or, b) change the directory to, say, /var/log?

I tried tinkering with the environment variable GP_LOG_DIR, I changed the Info.plist "-log" option to "false", but it only got rid of the ganttproject.log, not the java0.log.

Great app otherwise!

Thanks

chiyang10000:
Hello, I have found that modifying `/Applications/GanttProject.app/Contents/Java/Classes/logging.properties` helps to change the location of `java0.log`.

```
java.util.logging.FileHandler.pattern = /tmp/java%u.log
```
Hope this helps😃.

narekboghozian:
I also find this very annoying on MacOS, although I do like the app overall. 

My workaround was to run this command to hide the ganttproject.log: 

```
chflags hidden ~/ganttproject.log 
```

and for the java.log I changed:

```
java.util.logging.FileHandler.pattern = %h/java%u.log 
```

to:

```
java.util.logging.FileHandler.pattern = %h/.ganttproject.d/logs/java%u.log 
```

in:

```
/Applications/GanttProject.app/Contents/runtime/Contents/Home/conf/logging.properties 
```

and made a new `logs` folder in `.ganttproject.d`. Not sure if I disabled it but it isn't there anymore. During use the `ganttproject.log.lck` file is still present however. If there's a fix for that too I'd like to know.

PR #1393, opened 2017-04-07
fix issue #130

asarbs:


ganttproject/src/net/sourceforge/ganttproject/GanttProject.java
```diff
@@ -902,7 +902,7 @@ public static boolean main(String[] arg) {
       return false;
     }
     if (mainArgs.log && "auto".equals(mainArgs.logFile)) {
-      mainArgs.logFile = System.getProperty("user.home") + File.separator + "ganttproject.log";
+      mainArgs.logFile = System.getProperty("user.home")+ File.separator + ".ganttproject.d" + File.separator + "ganttproject.log";
```
dbarashev: This is likely to fail if directory doesn't exist, but even if it is fixed or doesn't fail, I don't think that we should create hidden directory `.ganttproject.d` on all systems.

When talking to Windows/Mac OSX users it is way more easier to tell them _find `ganttproject.log` file in their home directory_ than _find a hidden directory `.ganttproject.d` (you may need to enable hidden files showing in your file manager), then find `ganttproject.log` file in that directory_

Besides, issue #1309  requests chaning log directory to `/var/log/` or whatever user specifies in the command line, rather than to hidden directory.

asarbs: OK, then please scrap this pull request.

PR #1396, opened 2017-04-12
Import of Excel files directly into GanttProject

dbarashev:
@okusha Oleg, do you want to look at this?


okusha:
Have we lost changes in `i18n.properties` ? They were here: https://github.com/bardsoftware/ganttproject/pull/1396/commits/b51e7d55aa8cffe0cab7fe50cc27493ee61b5493#diff-07a6f95d9fbd985570cd24798f906018 but absent in this pull request

dbarashev:
Apparently our colleague @torkhov reverted i18n.properties along with changes in the Russian translation. Fixed.

okusha:
My next findings are:

1. Resource's load on a task does not survive the export-import process and is lost. So if some resource was assigned with 100.0 % load to a task it becomes 0.0 %. If this information is not preserved on purpose which I doubt then default load should be 100.0 instead of 0.0 I guess.
2. Default directory in import file chooser is not consistent with export. It's user home dir for export and current dir for import. This makes user to do a lot of extra clicks if several export-import cycles are needed.

I think all above behavior is not introduced by this pull request and have been sitting in the system before it. So if we consider these issues as bugs or feature requests then it LGTM from me.

BTW I can fix the findings myself if you want

PR #1419, opened 2017-06-02
A bunch of changes to add a second cost field for each task

bosmanoglu:
These changes are a first attempt to add a second Cost field for each task, replicating the features of the original cost field. 
What works:
- Saving and Loading *.gan files. 
- Calculation of total cost in Resources 
What doesn't work:
- Can't display Cost2 in Gantt Chart Tree View
- Messed-up the Cost label in Task> Resources. It says option.Group.task.cost2.label.
- The Cost2 Column in Resource view works but has a column title "tableColResourceCost2"
Everything else is untested so far.


biz.ganttproject.core/src/biz/ganttproject/core/model/task/TaskDefaultColumn.java
```diff
@@ -50,7 +50,8 @@
   ID(new ColumnList.ColumnStub("tpd10", null, false, -1, 20), Integer.class, "tableColID", Functions.NOT_EDITABLE),
   OUTLINE_NUMBER(new ColumnList.ColumnStub("tpd11", null, false, 4, 20), String.class, "tableColOutline", Functions.NOT_EDITABLE),
   COST(new ColumnList.ColumnStub("tpd12", null, false, -1, 20), Double.class, "tableColCost"),
-  RESOURCES(new ColumnList.ColumnStub("tpd13", null, false, -1, 20), String.class, "resources", Functions.NOT_EDITABLE);
+  COST2(new ColumnList.ColumnStub("tpd13", null, false, -1, 20), Double.class, "tableColCost"),
```
dbarashev: I can't test because CostAlgorithmImpl2 is missing, but here you're using the same name key `tableColCost` in your new column. That might be the reason of the troubles with the table.

ganttproject/src/net/sourceforge/ganttproject/gui/GanttDialogPerson.java
```diff
@@ -78,6 +81,7 @@ public GanttDialogPerson(CustomPropertyManager customPropertyManager, TaskManage
 
     ((GPAbstractOption)myTotalCostField).setWritable(false);
     myRateGroup = new GPOptionGroup("resourceRate", new GPOption[] {myStandardRateField, myTotalCostField});
+    myRate2Group = new GPOptionGroup("resourceRate", new GPOption[] {myStandardRate2Field, myTotalCost2Field});
```
dbarashev: The same normally applies to option groups: use different ids for different groups

bosmanoglu:
OK, I think I got those issues sorted out. Thanks so much for your help. 

I also got Cost2 to export with CSV. I tried adding "per resource (person) total cost" information for both Cost and Cost2 in the CSV export tables, but haven't been successful. I actually couldn't change the resource related part of the CSV table at all. 

Any ideas?

dbarashev:
You just need to check the new fields in the CSV settings.

bosmanoglu:
That was easy! OK the last thing (I think) I would like to do is to show the sum of "Custom Columns" for child tasks (e.g. such that the parent tasks shows the sum of children) in the Gantt Table. If you have any pointers on where to start let me know. I am hoping to look at it this weekend or so.

dbarashev:
What do you mean as "the sum of custom columns" ? Do you want to do calculations with arbitrary user-defined columns or with some known subset? Or just with one column?

PR #1424, opened 2017-06-07
Task table updated with resource unit

dejande:


ganttproject/src/net/sourceforge/ganttproject/GanttTreeTableModel.java
```diff
@@ -318,7 +318,10 @@ else if (t.isMilestone()) {
     	List<String> resources = Lists.transform(Arrays.asList(t.getAssignments()), new Function<ResourceAssignment, String>() {
 			@Override
 			public String apply(ResourceAssignment ra) {
-				return ra.getResource().getName();
+			  float load = ra.getLoad();
+			  String name = ra.getResource().getName();
+
+			  return (load == 100)?name:name.concat(String.format("[%d%%]", Math.round(load)));
```
dbarashev: Please include the name itself into the pattern (instead of concatenating) and extract the pattern into i18n.properties file so that it could be localized. I suggest using `taskColResources.value` as key. You may use placeholders in MessageFormat syntax in the i18n value, like this:
`taskColResources.value = {0} [{1}%]`

and pass string values of name and load into `GanttLanguage::formatText` method.

Also, please surround `?` and `:` with spaces.

PR #1442, opened 2017-07-24
Schch resource total load

schch:
Hi there,

based on the new feature which calculates the total cost of a resource I made a patch to calculate the total load of a resource. A feature I really missed so far.

Test and translation for DE included.

Please let me know of any problems which prevent the feature to be used for the next release, so I can fix them.

Regards and thanks for your work,
Christoph

PS: First I tried to fix a potential problem with the the calculation of the total cost in case of tasks with fixed costs. Hence the unnecessary commits as I never worked with git before. There is no bug with the calculation of the total cost.


ganttproject-tester/test/net/sourceforge/ganttproject/resource/TestResourceTotalLoad.java
```diff
@@ -0,0 +1,38 @@
+package net.sourceforge.ganttproject.resource;
```
dbarashev: Please add copyright notice:
```
/*
Copyright 2017 Christoph Schneider, BarD Software s.r.o

This file is part of GanttProject, an opensource project management tool.

GanttProject is free software: you can redistribute it and/or modify
it under the terms of the GNU General Public License as published by
 the Free Software Foundation, either version 3 of the License, or
 (at your option) any later version.

GanttProject is distributed in the hope that it will be useful,
but WITHOUT ANY WARRANTY; without even the implied warranty of
MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
GNU General Public License for more details.

You should have received a copy of the GNU General Public License
along with GanttProject.  If not, see <http://www.gnu.org/licenses/>.
*/
```

ganttproject-tester/test/net/sourceforge/ganttproject/resource/TestResourceTotalLoad.java
```diff
@@ -0,0 +1,38 @@
...
+
+import net.sourceforge.ganttproject.TestSetupHelper;
+import net.sourceforge.ganttproject.TestSetupHelper.TaskManagerBuilder;
+import net.sourceforge.ganttproject.task.Task;
+import net.sourceforge.ganttproject.test.task.TaskTestCase;
+
+public class TestResourceTotalLoad extends TaskTestCase {
+
+    public void testResourceTotalLoad() {
+	    TaskManagerBuilder builder = TestSetupHelper.newTaskManagerBuilder();
```
dbarashev: Please reindent this file with spaces (not tabs), 2 spaces indent size

schch:
OK, fixed.

PR #1448, opened 2017-08-04
Tkt 405 comment3 total work load

schch:
Hi there,

this adds the total load of a task (hierarchy) as described in comment 3 of issue 405 except for the part where it says work hours instead of days.

The load is calculated based on the resource assignments of a task and its subtasks similar to the total cost of a task. The value is read only and is visible at the same places as the total cost of a task.

Please let me know if you'd like any changes.

Cheers,
Christoph


dbarashev:
Hi Christoph,
I will look at this at the end of next week when I come back from vacation.

ganttproject/src/net/sourceforge/ganttproject/gui/taskproperties/TaskAllocationsPanel.java
```diff
@@ -78,7 +80,7 @@ public JPanel getComponent() {
     CommonPanel.setupComboBoxEditor(getTable().getColumnModel().getColumn(4), myRoleManager.getEnabledRoles());
 
     JPanel tablePanel = CommonPanel.createTableAndActions(myTable, myModel);
-    String layoutDef = "(ROW weight=1.0 (LEAF name=resources weight=0.5) (LEAF name=cost weight=0.5))";
+    String layoutDef = "(ROW weight=1.0 (LEAF name=resources weight=0.5) (COLUMN weight=0.5 (LEAF name=cost weight=0.5) (LEAF name=load weight=0.5)))";
```
dbarashev: Please change the weights of cost and load to 0.1 -- this will group them in the upper part of the column closer to each other

ganttproject/src/net/sourceforge/ganttproject/gui/taskproperties/TaskAllocationsPanel.java
```diff
@@ -122,6 +125,21 @@ public void changeValue(ChangeValueEvent event) {
     return result;
   }
 
+  private JComponent createLoadPanel() {
+    OptionsPageBuilder builder = new OptionsPageBuilder();
+
+    JPanel optionsPanel = new JPanel();
```
dbarashev: This code is either over- or under- engineered :)
`GPOption` and `GPOptionGroup` classes serve as UI-independent "properties", or as a model for UI components if you wish. `OptionsPageBuilder` can automatically build UI from options and groups, adds titles, makes them readonly depending on the model properties. If you use this machinery then it makes sense to utilize it completely and do not bother with UI components explicitly (and it is the recommended way, unless UI is more complex than property sheet -like)

In order to use it, remove all code at lines 130-140 and replace it with 
```
return builder.createGroupComponent(myLoadGroup);
```
Also, add the option instance:
```
  private final DefaultDoubleOption myLoadOption = new DefaultDoubleOption("taskProperties.load.value");
```
and pass it to `myLoadGroup` in the group constructor call.

You're almost done, the remaining thing is to set the value of `myLoadOption` from `myTask`

ganttproject/src/net/sourceforge/ganttproject/task/TaskImpl.java
```diff
@@ -1277,4 +1280,19 @@ public void setCalculated(boolean calculated) {
...
+    @Override
+    public Double getValue() {
+      return new LoadAlgorithmImpl().getCalculatedLoad(TaskImpl.this);
+    }
+
+  }
+
+  @Override
+  public Load getLoad() {
+	return myLoad;
```
dbarashev: Pls replace tab with 2 spaces

ganttproject/src/net/sourceforge/ganttproject/task/algorithm/LoadAlgorithmImpl.java
```diff
@@ -0,0 +1,44 @@
...
+import net.sourceforge.ganttproject.task.TaskContainmentHierarchyFacade;
+
+/**
+ * Algorithm for calculating task load
+ *
+ * @author schch (Christoph Schneider)
+ */
+public class LoadAlgorithmImpl {
+  public Double getCalculatedLoad(Task t) {
+    Double total = new Double(0.0);
```
dbarashev: Any sense to use primitive type while calculating? Not sure if it matters these days from performance point, but summing non-primitive Double's in the loops below must heavily and pointlessly use autoboxing and unboxing

ganttproject/src/net/sourceforge/ganttproject/task/TaskImpl.java
```diff
@@ -1277,4 +1280,19 @@ public void setCalculated(boolean calculated) {
   public Cost getCost() {
     return myCost;
   }
+
+  private class LoadImpl implements Load {
+
+    @Override
+    public Double getValue() {
+      return new LoadAlgorithmImpl().getCalculatedLoad(TaskImpl.this);
```
dbarashev: Out of curiosity, is it possible to cache the calculated result? If it is always calculated then we're going to use CPU heavily in big projects.

Probably a good place to store load, cache it and invalidate the cache is `ResourceAssignmentCollectionImpl` (field `myAssignments`) which is aware of assigned resources and their changes.

schch:
Hi Dmitry, getting back to it as soon as I can. Might take a couple of weeks.

dbarashev:
Hi, any news on this?

PR #1455, opened 2017-08-24
Schch fix completion percentage

schch:
Hi Dmitry,

it looks like the calculation of the completion percentage is inconsistent.

When the depth of a task hierarchy is greater than 1 and there are gaps between tasks the percentage of the tasks at least two levels above a leave are wrong.
completion-broken

super task_0 and super task_1 should have the same percentage like both instances of the nesteted tasks task_0 and task_3.

In addition to this a milestone counts as a day towards completion but has a duration of 0. This also seems to be inconsistent.

The solution is to calculate the completion using the recursively accumulated completed and total duration of all tasks which do not have nested tasks. For a given tasks this has to be done for the complete part of the hierarchy i.e. sub-tree the task is a part of.

This pull request fixes both issues and includes a new test.

Cheers,
Christoph


ganttproject/src/net/sourceforge/ganttproject/task/algorithm/RecalculateTaskCompletionPercentageAlgorithm.java
```diff
@@ -29,7 +29,7 @@ public void run() {
     }
     TaskContainmentHierarchyFacade facade = createContainmentFacade();
     for (Task t : facade.getTasksInDocumentOrder()) {
-      if (!facade.hasNestedTasks(t)) {
+      if (facade.getDepth(t) == 0) {
```
dbarashev: Since the algorithm is now going top-down, you can remove this loop completely and just take the root task from the facade.

ganttproject/src/net/sourceforge/ganttproject/task/algorithm/RecalculateTaskCompletionPercentageAlgorithm.java
```diff
@@ -43,29 +43,52 @@ public void run(Task task) {
     recalculateSupertaskCompletionPercentageBottomUp(task, facade);
   }
 
+  private class PropagatedDays {
+    public int myCompletedDays;
```
dbarashev: Pls make the class static and fields final

ganttproject/src/net/sourceforge/ganttproject/task/algorithm/RecalculateTaskCompletionPercentageAlgorithm.java
```diff
@@ -43,29 +43,52 @@ public void run(Task task) {
...
+    public long myPlannedDays;
+
+    PropagatedDays() {
```
dbarashev: You probably don't need this constructor

ganttproject/src/net/sourceforge/ganttproject/task/algorithm/RecalculateTaskCompletionPercentageAlgorithm.java
```diff
@@ -43,29 +43,52 @@ public void run(Task task) {
     recalculateSupertaskCompletionPercentageBottomUp(task, facade);
   }
 
+  private class PropagatedDays {
+    public int myCompletedDays;
+    public long myPlannedDays;
```
dbarashev: Why planned is long while completed is int? In the code, completed is like 100x planned so it should be at least wider

ganttproject/src/net/sourceforge/ganttproject/task/algorithm/RecalculateTaskCompletionPercentageAlgorithm.java
```diff
@@ -43,29 +43,52 @@ public void run(Task task) {
...
+      myPlannedDays = 0;
+    };
+
+    PropagatedDays(int completedDays, long plannedDays) {
+      myCompletedDays = completedDays;
+      myPlannedDays = plannedDays;
+    };
+  }
+
   private void recalculateSupertaskCompletionPercentageBottomUp(Task task, TaskContainmentHierarchyFacade facade) {
```
dbarashev: If you take the root task at lines 31-35, you can remove this method completely.

ganttproject/src/net/sourceforge/ganttproject/task/algorithm/RecalculateTaskCompletionPercentageAlgorithm.java
```diff
@@ -43,29 +43,52 @@ public void run(Task task) {
...
-        completedDays += nextDuration * next.getCompletionPercentage();
-        plannedDays += nextDuration;
+        PropagatedDays propagatedDays = recalculateSupertaskCompletionPercentage(next, facade);
+        completedDays += propagatedDays.myCompletedDays;
+        plannedDays += propagatedDays.myPlannedDays;
       }
       int completionPercentage = plannedDays == 0 ? 0 : (int) (completedDays / plannedDays);
       task.setCompletionPercentage(completionPercentage);
+    } else {
+      long nextDuration = task.getDuration().getLength();
```
dbarashev: I suggest returning `new PropagatedDays(nextDuration * task.getCompletionPercentage(), nextDuration)` right away and moving accumulating vars into the previous branch of if

ganttproject/src/net/sourceforge/ganttproject/task/algorithm/RecalculateTaskCompletionPercentageAlgorithm.java
```diff
@@ -43,29 +43,52 @@ public void run(Task task) {
     recalculateSupertaskCompletionPercentageBottomUp(task, facade);
   }
 
+  private class PropagatedDays {
```
dbarashev: Let's rename it to `SubtreeCompletion` ?

ganttproject-tester/test/net/sourceforge/ganttproject/test/task/TestTaskCompletionPercentage.java
```diff
@@ -124,6 +124,54 @@ public void testCompletionWithMilestones() {
...
+    Task supertask = taskManager.createTask();
+
+    Task supertask_0 = taskManager.createTask();
+    Task supertask_0_0 = taskManager.createTask();
+    Task supertask_0_1 = taskManager.createTask();
+    Task supertask_1 = taskManager.createTask();
+    Task supertask_1_0 = taskManager.createTask();
+    Task supertask_1_1 = taskManager.createTask();
+
+    supertask_0_0.setStart(CalendarFactory.createGanttCalendar(2000, 01, 10));
```
dbarashev: I recommend using `TaskBuilder` interface, like this:
```
Task task_0_0 = taskManager.newTaskBuilder()
    .withStartDate(TestSetupHelper.newMonday().getTime())
    .withDuration(taskManager.createLength(1))
    .withCompletion(100)
    .withParent(supertask_0)
    .build();
```

This will make the test code more readable. 

Please pay attention that in the code end date is start date + 1.

ganttproject-tester/test/net/sourceforge/ganttproject/test/task/TestTaskCompletionPercentage.java
```diff
@@ -124,6 +124,54 @@ public void testCompletionWithMilestones() {
...
+    supertask_1_1.setStart(CalendarFactory.createGanttCalendar(2000, 01, 14));
+    supertask_1_1.setEnd(CalendarFactory.createGanttCalendar(2000, 01, 14));
+
+    supertask_1_0.move(supertask_1);
+    supertask_1_1.move(supertask_1);
+
+    supertask_0.move(supertask);
+    supertask_1.move(supertask);
+
+    supertask.move(project);
```
dbarashev: Project has to be moved under `taskManager.getRootTask()` as well.

ganttproject-tester/test/net/sourceforge/ganttproject/test/task/TestTaskCompletionPercentage.java
```diff
@@ -124,6 +124,54 @@ public void testCompletionWithMilestones() {
...
+    //
+    RecalculateTaskCompletionPercentageAlgorithm alg = taskManager.getAlgorithmCollection().getRecalculateTaskCompletionPercentageAlgorithm();
+    alg.run(project);
```
dbarashev: Maybe let's just run the algorithm without arguments, applying to the whole project?

ganttproject/src/net/sourceforge/ganttproject/task/algorithm/RecalculateTaskCompletionPercentageAlgorithm.java
schch: Ah, ok. So there is always a single root task no matter of the project structure?

ganttproject/src/net/sourceforge/ganttproject/task/algorithm/RecalculateTaskCompletionPercentageAlgorithm.java
schch: Ok, but then the implementation of `void run(Task task)` has to be changed too or its invocations need to be replaced by` run()`.

ganttproject/src/net/sourceforge/ganttproject/task/algorithm/RecalculateTaskCompletionPercentageAlgorithm.java
schch: Any particular reason for this?
This way there have to be two return statements, one within each branch of the if-statement which leaves the method without an obvious return statement.

ganttproject-tester/test/net/sourceforge/ganttproject/test/task/TestTaskCompletionPercentage.java
schch: Nice. I just used the available code within the existing tests.

ganttproject/src/net/sourceforge/ganttproject/task/algorithm/RecalculateTaskCompletionPercentageAlgorithm.java
dbarashev: Just to have less mutable vars. In fact, I would even remove else like below but that doesn't matter too much to me :)

```
if (nestedTasks.length == 0) {
  return new SubtreeCompletion....
}
long completedDays...
// other code from current if > 0 block

PR #1456, opened 2017-08-31
improving format of exported excel

schch:
The exported XLS file uses strings for every exported value and a single sheet for task an resources. In order to use the file for doing some calculations you have to change the columns by hand to have the correct type and format.

I moved the formatting of exported values to the SpreadsheetWriter by overloading the print method. I also extended the interface to be able to use more than one sheet. Now the created XLS uses date and number formats for the exported values where appropriate and separate sheets for tasks and resources. The re-import of an export to XLS with new and old format is still possible. CSV im-/export remains unchanged.

This also fixes a  bug where empty (custom) properties at the end of a line couldn't be read as the mapping of headers to column number was missing.

PS: When I re-imported an export of the example project the cost and coordinator properties where imported as custom properties in addition to their actual target.

### DIFF
--- a/ganttproject/src/biz/ganttproject/impex/csv/CsvRecordImpl.java
+++ b/ganttproject/src/biz/ganttproject/impex/csv/CsvRecordImpl.java
@@ -35,7 +35,7 @@ class CsvRecordImpl implements SpreadsheetRecord {
 
   @Override
   public String get(String name) {
-    return myRecord.get(name);
+    return (isSet(name)) ? myRecord.get(name) : new String();
   }
 
   @Override
--- a/ganttproject/src/biz/ganttproject/impex/csv/XlsReaderImpl.java
+++ b/ganttproject/src/biz/ganttproject/impex/csv/XlsReaderImpl.java
@@ -21,9 +21,13 @@ package biz.ganttproject.impex.csv;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Iterators;
 import com.google.common.collect.Lists;
+
+import net.sourceforge.ganttproject.language.GanttLanguage;
+
 import org.apache.poi.hssf.usermodel.HSSFWorkbook;
-import org.apache.poi.ss.usermodel.Cell;
+import org.apache.poi.hssf.usermodel.HSSFDateUtil;
 import org.apache.poi.ss.usermodel.Row;
+import org.apache.poi.ss.usermodel.Sheet;
 import org.apache.poi.ss.usermodel.Workbook;
 
 import java.io.IOException;
@@ -32,6 +36,7 @@ import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.text.SimpleDateFormat;
 
 /**
  * @author torkhov
@@ -40,10 +45,12 @@ class XlsReaderImpl implements SpreadsheetReader {
 
   private final Workbook myBook;
   private final Map<String, Integer> myHeaders;
+  private final SimpleDateFormat myDateFormat;
 
   XlsReaderImpl(InputStream is, List<String> columnHeaders) throws IOException {
     myBook = new HSSFWorkbook(is);
     myHeaders = initializeHeader(columnHeaders);
+    myDateFormat = GanttLanguage.getInstance().getShortDateFormat();
   }
 
   @Override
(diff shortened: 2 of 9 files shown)


ganttproject/src/biz/ganttproject/impex/csv/CsvRecordImpl.java
```diff
@@ -35,7 +35,7 @@
 
   @Override
   public String get(String name) {
-    return myRecord.get(name);
+    return (isSet(name)) ? myRecord.get(name) : new String();
```
dbarashev: just return `""` instead of `new String()`

ganttproject/src/biz/ganttproject/impex/csv/XlsReaderImpl.java
```diff
@@ -53,11 +60,50 @@ public void close() throws IOException {
...
-    return Iterators.transform(myBook.getSheetAt(0).iterator(), (input) -> new XlsRecordImpl(getCellValues(input), myHeaders));
+    Iterator<SpreadsheetRecord> iterator = null;
+
+    if(myBook.getNumberOfSheets() > 1) {
+      Iterator<SpreadsheetRecord> it = null;
+      for (int i = 0; i < myBook.getNumberOfSheets(); ++i) {
+        Sheet sheet = myBook.getSheetAt(i);
+        if (i < myBook.getNumberOfSheets() - 1) {
+          int lastRow = sheet.getPhysicalNumberOfRows();
+          sheet.createRow(lastRow + 1);
```
dbarashev: This looks a little bit weird: why do we create rows when reading? Even if it makes no changes on the disk (I believe it doesn't) it worth adding a comment about the purpose of these two rows.

ganttproject/src/biz/ganttproject/impex/csv/XlsReaderImpl.java
```diff
@@ -53,11 +60,50 @@ public void close() throws IOException {
...
+          sheet.createRow(lastRow + 2);
+        }
+        it = Iterators.transform(sheet.iterator(), (input) -> new XlsRecordImpl(getCellValues(input), myHeaders));
+        iterator = (iterator == null) ? it : Iterators.concat(iterator, it);
+      }
+    }
+    else {
```
dbarashev: It seems that you can safely remove `else` block and `if` condition and just do what is written in the true-case block no matter if you have just 1 sheet or more than 1.

ganttproject/src/biz/ganttproject/impex/csv/XlsReaderImpl.java
```diff
@@ -53,11 +60,50 @@ public void close() throws IOException {
...
+      iterator = Iterators.transform(myBook.getSheetAt(0).iterator(), (input) -> new XlsRecordImpl(getCellValues(input), myHeaders));
+    }
+
+    return iterator;
   }
 
   private List<String> getCellValues(Row row) {
-    return Lists.newArrayList(Iterables.transform(row, Cell::getStringCellValue));
+    return Lists.newArrayList(Iterables.transform(row,
+            (input) -> {
```
dbarashev: Looks like too much indent here. We use 4 spaces indent for continuations, but here the following formatting might be even better:
```
  return Lists.newArrayList(Iterables.transform(row, (input) -> {
    switch
  }));

File: ganttproject/src/biz/ganttproject/impex/csv/XlsReaderImpl.java [hunk 5]
```diff
@@ -53,11 +60,50 @@ public void close() throws IOException {
...
+              switch (input.getCellTypeEnum()) {
+                case NUMERIC: {
+                  if (HSSFDateUtil.isCellDateFormatted(input)) {
+                    return myDateFormat.format(input.getDateCellValue());
+                  }
+                  else {
+                    if (input.getCellStyle().getDataFormat() == (short)1) {
```
dbarashev: This looks as if it produces the same result as subsequent else, no? Can we just call `String.valueOf(input.getNumericCellValue())` no matter what data format is? Or even just call `getStringCellValue` in all cases when `isCellDateFormatted` returns false?

ganttproject/src/biz/ganttproject/impex/csv/XlsRecordImpl.java
```diff
@@ -44,7 +44,7 @@ public String get(String name) {
     if (index == null) {
       throw new IllegalArgumentException(String.format("Mapping for %s not found, expected one of %s", name, myMapping.keySet()));
     }
-    return myValues.get(index);
+    return (myValues.size() <= index) ? new String() : myValues.get(index);
```
dbarashev: s/new String()/"" ?

ganttproject/src/biz/ganttproject/impex/csv/XlsWriterImpl.java
```diff
@@ -20,45 +20,65 @@
...
+    myDoubleStyle.setDataFormat((short) 2);
+
+    myCurrentSheet = myWorkbook.createSheet(initialSheetName);
   }
 
   @Override
   public void print(String value) throws IOException {
-    if (myCurrentRow == null) {
-      createNewRow();
+    if (value != null) {
```
dbarashev: is it possible that value is nullable? What does it mean from exporter's point of view? If it is empty cell then we should probably add a cell anyway, just don't fill it with value.

ganttproject/src/biz/ganttproject/impex/csv/XlsReaderImpl.java
schch: That is because of the general CSV-Reader which expects to have two empty lines between tables e.g. tasks and ressources.

PR #1494, opened 2018-02-21
Prettyfying the search results when the count of results is less than 9.

dkelly-sighter:
Feel free to say this is worthless or that 9 belongs as a constant somewhere instead of being a magic number.

### DIFF
--- a/ganttproject/src/net/sourceforge/ganttproject/search/SearchUiImpl.java
+++ b/ganttproject/src/net/sourceforge/ganttproject/search/SearchUiImpl.java
@@ -115,6 +115,8 @@ public class SearchUiImpl implements SearchUi {
       int searchResultLength = searchResults.length;
       if (searchResultLength < 9) {
         list.setVisibleRowCount(searchResultLength);
+      } else {
+        list.setVisibleRowCount(10);
       }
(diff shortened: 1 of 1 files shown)


ganttproject/src/net/sourceforge/ganttproject/search/SearchUiImpl.java
```diff
@@ -110,7 +110,12 @@ public void accept(final List<SearchResult<?>> results) {
       if (results.isEmpty()) {
         return;
       }
-      list.setListData(results.toArray(new SearchResult[0]));
+      SearchResult[] searchResults = results.toArray(new SearchResult[0]);
+      list.setListData(searchResults);
+      int searchResultLength = searchResults.length;
+      if (searchResultLength < 9) {
+        list.setVisibleRowCount(searchResultLength);
```
dbarashev: Please add else block which would set visible row count to 10 if we have enough results. Otherwise the dropdown shrinks to very small height after the first search with few results.

dbarashev:
This totally makes sense, thanks!

If you're interested in improving search function, I can drop a few more ideas on that.  Ping me here or email to dbarashev@ganttproject.biz or join chat in Slack (I'll send you an invitation)

PR #1512, opened 2018-04-02
Added update detection and update dialog

ALapinskyi:


ganttproject/src/net/sourceforge/ganttproject/client/RssFeedChecker.java
```diff
@@ -228,6 +236,14 @@ public void run() {
     };
   }
 
+  private void createUpdateDialog(String content) {
+    RssUpdate update = parser.parseUpdate(content);
+    if(update != null) {
+      Thread updateDialogThread = new Thread(() -> myUiFacade.showUpdateDialog(update));
```
dbarashev: I don't think that creating Thread is necessary because UI must be shown in Swing thread and subsequent download must run NOT in Swing thread. 
`SwingUtilities.invokeLater` should do the trick.

ganttproject/src/net/sourceforge/ganttproject/client/RssParser.java
```diff
@@ -84,6 +85,18 @@ public RssFeed parse(InputStream inputStream, Date lastCheckDate) {
     return result;
   }
 
+  public RssUpdate parseUpdate(String content){
+    if(!StringUtils.isEmptyOrNull(content)) {
+      content = content.replaceAll("</?div>(\\n)?", "");
```
dbarashev: It is not that easy because, although the content is HTML, it is 1) escaped HTML and 2) there are attributes inside div

With respect to 1), you can unescape the content with `com.google.common.xml.XmlEscapers#xmlAttributeEscaper`.

With respect to 2), it seems that we'll have to build this feed on our own anyway, so we'll filter out the wrapping `div`. You may think that after unescaping it will be exactly like 
```
VERSION
LINK
SOME HTML
```

You can find an example  here: https://www.dropbox.com/s/6dnpdw3evhddcad/update.rss?dl=1 
It is a copy of our current feed plus update entry which is formatted in the way described above.

You can temporarily replace net.sourceforge.ganttproject.client.RssFeedChecker#RSS_URL value with that URL

ganttproject/src/net/sourceforge/ganttproject/client/RssParser.java
```diff
@@ -142,6 +155,21 @@ private void addItem(RssFeed result, Node item, Date lastCheckDate) throws XPath
...
+    result.addItem(title, body, isUpdate);
+  }
+
+  private boolean isUpdateItem(Node item) throws XPathExpressionException {
+    NodeList categories = (NodeList) getXPath("atom:category").evaluate(item, XPathConstants.NODESET);
+    for (int i = 0; i < categories.getLength(); i++) {
+      Element elCategory = (Element) categories.item(i);
+      String category = elCategory.getAttribute("term");
+      if (!Strings.isNullOrEmpty(category)) {
+        if(category.equals("update")){
```
dbarashev: Please add a space after `if` and before `{`

ganttproject/src/net/sourceforge/ganttproject/client/RssUpdate.java
```diff
@@ -0,0 +1,44 @@
...
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with GanttProject.  If not, see <http://www.gnu.org/licenses/>.
+*/
+package net.sourceforge.ganttproject.client;
+
+public class RssUpdate {
+
+    private String version;
```
dbarashev: Please do follow the [code style](https://docs.ganttproject.biz/development/#_hack)
Two spaces, `my` prefix (and `this` is not required). Also, these fields can be made final

ganttproject/src/net/sourceforge/ganttproject/client/RssParser.java
```diff
@@ -84,6 +85,18 @@ public RssFeed parse(InputStream inputStream, Date lastCheckDate) {
     return result;
   }
 
+  public RssUpdate parseUpdate(String content){
+    if(!StringUtils.isEmptyOrNull(content)) {
```
dbarashev: Please add a space after `if`

ganttproject/src/net/sourceforge/ganttproject/client/RssFeedChecker.java
```diff
@@ -228,6 +236,14 @@ public void run() {
     };
   }
 
+  private void createUpdateDialog(String content) {
+    RssUpdate update = parser.parseUpdate(content);
+    if(update != null) {
```
dbarashev: Space after `if`

ganttproject/src/net/sourceforge/ganttproject/gui/UIFacade.java
```diff
@@ -154,4 +155,6 @@
   GPOptionGroup[] getOptions();
 
   void addOnUpdateComponentTreeUi(Runnable callback);
+
+  void showUpdateDialog(RssUpdate update);
```
dbarashev: I think that this UIFacade layer is not necessary. You can just call `UpdateDialog` from `RssFeedChecker`

ganttproject/data/resources/language/i18n.properties
```diff
@@ -645,6 +645,7 @@ typeClass = Type
 undo = Undo
 unit = Unit
 unitsPerDay = Maximum units per day
+updateAvailable=An updated GanttProject is available! Install version {0} now?
```
dbarashev: pls surround `=` with spaces

dbarashev:
Please write a simple unit test for the code which extract update item. It seems to be the most error prone in this pull request.

ganttproject/src/net/sourceforge/ganttproject/client/RssParser.java
ALapinskyi: com.google.common.xml.XmlEscapers#xmlAttributeEscaper. escape also html tags inside the string
Maybe I should just remove wrapped div but not with regex that may also remove other div tags?

dbarashev: You need to unescape html tags anyway to pass HTML to the editor pane.

PR #1513, opened 2018-04-03
Added downloader for updates

ALapinskyi:


ganttproject/src/net/sourceforge/ganttproject/gui/update/RestartDialog.java
```diff
@@ -0,0 +1,62 @@
...
+    JPanel panel = new JPanel();
+    panel.add(label);
+
+    OkAction okAction = new OkAction() {
+      @Override
+      public void actionPerformed(ActionEvent e) {
+        ScheduledExecutorService schedulerExecutor = Executors.newScheduledThreadPool(2);
+        Callable<Process> callable = () -> {
+          ProcessBuilder builder = new ProcessBuilder(
+                  SystemUtils.IS_OS_WINDOWS? "cmd /c start \"\" ganttproject.bat" : "ganttproject");
```
ALapinskyi: I'm cannot be at 100% sure that is works on Windows, but Google says so..

ganttproject/src/net/sourceforge/ganttproject/DownloadWorker.java
```diff
@@ -0,0 +1,150 @@
...
+  private UIFacade myUiFacade;
+
+  public DownloadWorker(UIFacade uiFacade, String downloadURL) {
+    myUiFacade = uiFacade;
+    myDownloadURL = downloadURL;
+  }
+
+  @Override
+  protected void process(List<Integer> chunks) {
+    myUiFacade.setDownloadProgress(chunks.get(chunks.size() - 1));
```
dbarashev: I suggest using UIFacade.setStatusText for this purpose (construct the string to be shown right here). Currently that method has no implementation, but it is trivial to implement.

ganttproject/src/net/sourceforge/ganttproject/gui/update/RestartDialog.java
```diff
@@ -0,0 +1,62 @@
...
+          return builder.start();
+        };
+        FutureTask<Process> futureTask = new FutureTask<>(callable);
+        schedulerExecutor.submit(futureTask);
+
+        System.exit(0);
+      }
+    };
+
+    UIFacade.Dialog dlg = uiFacade.createDialog(panel, new Action[]{okAction, CancelAction.EMPTY}, "");
```
dbarashev: I think you can use UIFacade.showConfirmationDialog or showOptionDialog here.

ganttproject/src/net/sourceforge/ganttproject/gui/update/RestartDialog.java
dbarashev: oh wow :) 
In fact, Eclipsito can detect when GanttProject closes its UI and relaunch, but that needs some polishing and I'll do it later. 

Besides, user may have unsaved changes and if we just exit then he will not be happy (even if he confirmed restart). 
Perhaps a good choice is to emulate window closing:
`uiFacade.getMainFrame().dispatchEvent(new WindowEvent(frame, WindowEvent.WINDOW_CLOSING));`

ganttproject/src/net/sourceforge/ganttproject/DownloadWorker.java
```diff
@@ -0,0 +1,150 @@
...
+      long totalBytesRead = 0;
+      int percentCompleted = 0;
+      long fileSize = urlConnection.getContentLength();
+
+      while ((bytesRead = inputStream.read(buffer)) != -1) {
+        outputStream.write(buffer, 0, bytesRead);
+        totalBytesRead += bytesRead;
+        percentCompleted = (int) (totalBytesRead * 100 / fileSize);
+
+        publish(percentCompleted);
```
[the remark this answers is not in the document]
ALapinskyi: "publish - process" it this situation is almost the same
"Publish" sends to GUI thread information that "process" processes
do you think that there is a reason refactor it to setProgress?

ganttproject/src/net/sourceforge/ganttproject/DownloadWorker.java
```diff
@@ -0,0 +1,150 @@
...
+      ex.printStackTrace();
+    }
+  }
+
+  private String getExtractionFolder() {
+    String projectPath = System.getProperty("user.dir");
+    if (new File(projectPath).canWrite()) {
+      return projectPath + PLUGINS_FOLDER;
+    } else {
+      return System.getProperty("user.home") + "/.ganttproject.d" + PLUGINS_FOLDER;
```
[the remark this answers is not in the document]
ALapinskyi: "home/username" + "/.ganttproject.d" + "/plugins"
where I need separator?

ganttproject/src/net/sourceforge/ganttproject/DownloadWorker.java
ALapinskyi: UIFacade.setStatusText is already used from other methods. And if I'll implement this method - it may show also other information in status bar. I think that it may create some issues with unexpected information on status bar..

ganttproject/src/net/sourceforge/ganttproject/gui/update/RestartDialog.java
ALapinskyi: it I had know it before...
So I need to write just?
uiFacade.getMainFrame().dispatchEvent(new WindowEvent(frame, WindowEvent.WINDOW_CLOSING));

ganttproject/src/net/sourceforge/ganttproject/DownloadWorker.java
```diff
@@ -0,0 +1,150 @@
...
+        fos.close();
+        zis.closeEntry();
+      }
+    } catch (IOException ex) {
+      ex.printStackTrace();
+    }
+  }
+
+  private String getExtractionFolder() {
+    String projectPath = System.getProperty("user.dir");
```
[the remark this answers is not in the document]
ALapinskyi: Hm, for me `Object.class.getResource("/plugins")` or Object.class.getResource("/plugins-2.8.6") return null
Maybe by default this folder isn't configured as resource location

ganttproject/src/net/sourceforge/ganttproject/gui/update/RestartDialog.java
dbarashev: yes, and if that causes window closing then it is okay. Another option is to call `GanttProject.quitApplication` but it is not exposed in any interfaces, and passing GanttProject instance here is a bit excessive.

ganttproject/src/net/sourceforge/ganttproject/DownloadWorker.java
dbarashev: Thanks for pointing this out. It appears that most of those usages are pointless, I'll remove them.

ganttproject/src/net/sourceforge/ganttproject/DownloadWorker.java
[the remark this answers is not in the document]
dbarashev: instead of `/`  I guess. It is going to be `\` on Windows

ganttproject/src/net/sourceforge/ganttproject/DownloadWorker.java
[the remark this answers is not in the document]
ALapinskyi: ok. but how do you want to see this part? I can implement with both approaches

ganttproject/src/net/sourceforge/ganttproject/DownloadWorker.java
ALapinskyi: In this case I'm able to use UIFacade.setStatusText  safely?

ganttproject/src/net/sourceforge/ganttproject/DownloadWorker.java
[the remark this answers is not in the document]
ALapinskyi: no conclusions?

Issue #1521, opened 2018-04-10
Keyboard shortcut to expand/collapse task nodes is broken


dbarashev:
It seems that Ctrl+Space which is supposed to expand/collapse task node doesn't work.

PR #1522, opened 2018-04-11
Rewritten key events processing in the tables. 

dbarashev:
This should make actions which create new rows in a table and start cell editing great again and should make Ctrl+Space working again.


dbarashev:
Update issue #1521

PR #1568, opened 2018-06-19
Sign-in pane with transition to browser pane

dbarashev:
* Sign in pane with working "Sign In" buton which opens a new browser tab to sign in to GP Cloud and receives a callback HTTP request with auth token on successful sign in
* Browser pane stub which is shown on receiving auth token


dbarashev:
#1562

Issue #1569, opened 2018-06-20
Fetch user teams from GP cloud 


dbarashev:
Use auth token, fetch the list of teams from GP cloud and add them to the file browser view

PR #1570, opened 2018-06-21
Show user teams in GP Cloud Storage UI

dbarashev:
This change makes the UI of GanttProject Cloud storage mostly matching the UI of WebDAV storage. They both use the same UI layout and controls. Currently we can only show the teams where current use is a member of.


dbarashev:
#1569

Issue #1598, opened 2018-09-13
NullPointerException while input url/username/password in webdav setting


ArayChou:
GanttProject 2.8.9 Pilsen (build 2335)
java version "10" 2018-03-20
Microsoft Windows [Version 10.0.17134.228]

Hello, I download the zip from https://codeload.github.com/bardsoftware/ganttproject/zip/ganttproject-2.8.8, unzip it to a folder, and double click ganttproject.bat to lunch the application. everything works fine. 

But I can NOT change/add webdav settings.  When I try to input URL/Username/Password in the WebDAV setting page. only one latter is showed in the input box.  and got error log as this:
(please check the whole error log file: https://github.com/ArayChou/static/blob/master/ganttproject.log )

Sep 13, 2018 8:26:14 AM net.sourceforge.ganttproject.GPLogger logToLogger
WARNING: null
java.lang.NullPointerException
	at net.sourceforge.ganttproject.document.webdav.WebDavOptionPageProvider$2.changeValue(WebDavOptionPageProvider.java:93)
	at biz.ganttproject.core.option.GPAbstractOption$Listeners.fire(GPAbstractOption.java:214)
	at biz.ganttproject.core.option.GPAbstractOption.fireChangeValueEvent(GPAbstractOption.java:130)
	at biz.ganttproject.core.option.GPAbstractOption.resetValue(GPAbstractOption.java:96)
	at biz.ganttproject.core.option.GPAbstractOption.setValue(GPAbstractOption.java:80)
	at net.sourceforge.ganttproject.gui.options.OptionsPageBuilder$14.saveValue(OptionsPageBuilder.java:367)
	at net.sourceforge.ganttproject.gui.options.OptionsPageBuilder$14.insertUpdate(OptionsPageBuilder.java:376)
	at java.desktop/javax.swing.text.AbstractDocument.fireInsertUpdate(Unknown Source)
	at java.desktop/javax.swing.text.AbstractDocument.handleInsertString(Unknown Source)
	at java.desktop/javax.swing.text.AbstractDocument.insertString(Unknown Source)
	at java.desktop/javax.swing.text.PlainDocument.insertString(Unknown Source)
	at java.desktop/javax.swing.text.AbstractDocument.replace(Unknown Source)
	at java.desktop/javax.swing.text.JTextComponent.replaceSelection(Unknown Source)
[...]

PR #1673, opened 2019-07-12
Avoid NPE when entering url/username/password in webdav settings.

juananpe:
Refer to issue #1598

This patch just avoids the NullPointerException that arises when the user types anything in the WebDav settings (url, username, pass) before adding a new server. The usability of this panel should be improved, though. For example, as mentioned, disabling WebDav settings input controls when there isn't a selected server.


dbarashev:
LGTM
Thanks for fixing. WebDAV servers setup refactoring is in progress. The plan is to replace this UI with the new one in "My Projects" dialog.

PR #1677, opened 2019-07-15
Show cost with 2 decimals

### LINKED ISSUE #1659: Option to change cost display format
One may want to show cost zero-padded or with higher decimal precision. 
See https://help.ganttproject.biz/t/cost-column-0-00/1332

oihanealbizuri:
Fixes #1659

### DIFF
--- a/ganttproject/src/net/sourceforge/ganttproject/GPTreeTableBase.java
+++ b/ganttproject/src/net/sourceforge/ganttproject/GPTreeTableBase.java
@@ -58,6 +58,7 @@ import javax.swing.event.TableColumnModelEvent;
 import javax.swing.event.TableColumnModelListener;
 import javax.swing.event.TreeExpansionEvent;
 import javax.swing.event.TreeExpansionListener;
+import javax.swing.table.DefaultTableCellRenderer;
 import javax.swing.table.JTableHeader;
 import javax.swing.table.TableCellEditor;
 import javax.swing.table.TableCellRenderer;
@@ -70,6 +71,7 @@ import java.awt.event.AdjustmentListener;
 import java.awt.event.KeyEvent;
 import java.awt.event.MouseAdapter;
 import java.awt.event.MouseEvent;
+import java.text.DecimalFormat;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


ganttproject/src/net/sourceforge/ganttproject/GPTreeTableBase.java
```diff
@@ -96,6 +98,15 @@ public void applyComponentOrientation(ComponentOrientation o) {
     }
   };
 
+  class DecimalRenderer extends DefaultTableCellRenderer {
+    private final DecimalFormat myFormatter = new DecimalFormat("#0.00");
```
dbarashev: The format of representing currency values is locale-specific. You want to use [NumberFormat::getCurrencyInstance](https://docs.oracle.com/javase/8/docs/api/java/text/NumberFormat.html#getCurrencyInstance-java.util.Locale-) and pass the currently selected locale. `GanttLanguage` is responsible for such things (and it stores the current locale)

dbarashev:
Thanks for the pull request. However, it needs to take all details into account. See my comments below.

ganttproject/src/net/sourceforge/ganttproject/GPTreeTableBase.java
```diff
@@ -905,12 +911,20 @@ TableHeaderUiFacadeImpl getTableHeaderUiFacade() {
   protected TableColumnExt newTableColumnExt(int modelIndex) {
     TableColumnExt result = new TableColumnExt(modelIndex);
     Class<?> columnClass = getTreeTableModel().getColumnClass(modelIndex);
-    TableCellRenderer renderer = createCellRenderer(columnClass);
+    String columnName = getTreeTableModel().getColumnName(modelIndex).toLowerCase();
+    TableCellRenderer renderer;
+    Boolean costColumn = false;
+    if (Double.class.equals(columnClass) && columnName.equals("cost")) {
```
dbarashev: This will work only if your system locale is English. Should it be some other locale, column name will not be "cost".

You can find TaskDefaultColumn instance by model index: `TaskDefaultColumn.values()[modelIndex]`

ganttproject/src/net/sourceforge/ganttproject/GPTreeTableBase.java
```diff
@@ -905,12 +911,20 @@ TableHeaderUiFacadeImpl getTableHeaderUiFacade() {
   protected TableColumnExt newTableColumnExt(int modelIndex) {
     TableColumnExt result = new TableColumnExt(modelIndex);
     Class<?> columnClass = getTreeTableModel().getColumnClass(modelIndex);
-    TableCellRenderer renderer = createCellRenderer(columnClass);
+    String columnName = getTreeTableModel().getColumnName(modelIndex).toLowerCase();
```
dbarashev: This class is abstract and is shared between Gantt chart and Resource chart. Please move all checks which are related to tasks into GanttTreeTable

PR #1678, opened 2019-07-15
Fixes #1667

### LINKED ISSUE #1667: Remember the last import folder
When importing files into GanttProject, import file chooser always opens at the application install folder. It does not make too much sense because such folders are normally read-only and/or just not used for project storage.

We could remember the last import folder and open it by default in the file chooser. Export already works that way.

Original discussion on the support portal:
https://help.ganttproject.biz/t/change-home-folder-setting/1441

Anaitz98:
Remember the last import folder and open it by default in the file chooser

### DIFF
--- a/ganttproject/src/net/sourceforge/ganttproject/importer/FileChooserPage.java
+++ b/ganttproject/src/net/sourceforge/ganttproject/importer/FileChooserPage.java
@@ -72,6 +72,9 @@ class FileChooserPage extends AbstractFileChooserPage {
       return;
     }
     myImporter.setFile(file);
+    if(file != null) {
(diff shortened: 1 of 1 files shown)


ganttproject/src/net/sourceforge/ganttproject/importer/FileChooserPage.java
```diff
@@ -72,6 +72,9 @@ protected void setFile(File file) {
       return;
     }
     myImporter.setFile(file);
+    if(file != null) {
```
dbarashev:
```suggestion
    if (file != null) {
```

PR #1691, opened 2019-11-07
Add french holidays for 2020 and 2021

Airthee:


dbarashev:
Thanks for contributing!

PR #1692, opened 2019-11-07
Allow keyboard shortcuts for link and unlink actions

urtzipu:
This changes allows to use keyboard shortcuts for link and unlink actions


biz.ganttproject.app.libs/build.gradle
```diff
@@ -2,6 +2,7 @@ buildscript {
     ext.kotlin_version = '1.3.+'
     repositories {
         mavenCentral()
+        maven { url 'https://dl.bintray.com/kotlin/kotlin-eap'}
```
dbarashev: This issue is already fixed in master. You need to git merge master into this branch.

PR #1696, opened 2019-11-18
Implement h2 manager that is able to run sql files

anstkras:


com.bardsoftware.constable/src/main/kotlin/com/bardsoftware/constable/H2Manager.kt
```diff
@@ -0,0 +1,24 @@
...
+import org.h2.tools.RunScript
+import java.io.FileReader
+import java.sql.DriverManager
+import java.sql.SQLException
+import java.util.logging.Level
+import java.util.logging.Logger
+
+class H2Manager {
+    companion object {
+        private val logger = Logger.getLogger(H2Manager::class.java.name)
```
dbarashev: You can just move this logger to the file level.

com.bardsoftware.constable/src/main/kotlin/com/bardsoftware/constable/Main.kt
```diff
@@ -0,0 +1,14 @@
+package com.bardsoftware.constable
+
+import java.io.File
+
+fun main() {
```
dbarashev: Pls add application plugin to `build.gradle` so that it could be launched with `gradle run`. You'll also need to add `settings.gradle`, otherwise it will not run in this GanttProject multi-project build environment

com.bardsoftware.constable/src/main/kotlin/com/bardsoftware/constable/Main.kt
```diff
@@ -0,0 +1,14 @@
...
+    println("Enter path to sql file with database schema to load")
+    val path = readLine()!!
```
dbarashev: Maybe replace it with the command line argument? 
There is a choice of nice libs for that:
* https://ajalt.github.io/clikt/
* https://github.com/xenomachina/kotlin-argparser
* https://github.com/Kotlin/kotlinx.cli
* http://jcommander.org/

Wrt !!, I understand it is a corner case when the result of `readLine` call becomes null, but anyway, the canonical Kotlin null-safe way is like this:
```
readLine()?.let { path ->
  // Here path is not null
  if (!File(path).exists())
....
}
```

or like this:

```
val path = readLine() ?: return
// Here path is not null. If readLine returns null we return as well.
```

com.bardsoftware.constable/src/main/kotlin/com/bardsoftware/constable/H2Manager.kt
anstkras: Done

com.bardsoftware.constable/src/main/kotlin/com/bardsoftware/constable/Main.kt
anstkras: Done

PR #1701, opened 2019-12-11
Use H2 to store database on file system

anstkras:


com.bardsoftware.constable/client/src/main/kotlin/com/bardsoftware/constable/client/ConstableClient.kt
```diff
@@ -0,0 +1,24 @@
...
+import com.github.ajalt.clikt.core.CliktCommand
+import com.github.ajalt.clikt.parameters.arguments.argument
+import java.sql.SQLException
+import java.util.logging.Level
+import java.util.logging.Logger
+
+private val logger = Logger.getLogger(H2Manager::class.java.name)
+
+class ConstableClient : CliktCommand() {
+    private val path by argument(help = "Path to database file")
```
dbarashev: Will `file()`[https://ajalt.github.io/clikt/api/clikt/com.github.ajalt.clikt.parameters.types/file/] method work for this (to make sure it is really a file) ?

dbarashev: Also, this database file seems to be unused. I mean, you can run `SELECT 2+2` query even when there is no file at all. I understand it is just the first baby step, but may be pass the command in another command line argument, execute it and save the file when command is comleted? So that we knew that our file is actually used and keeps the database state.

anstkras: Actually, path is path + database name as it is in URL: `jdbc:h2:[<path>]<databaseName>`, it could be splitted into 2 arguments.

com.bardsoftware.constable/client/src/main/kotlin/com/bardsoftware/constable/client/ConstableClient.kt
```diff
@@ -0,0 +1,33 @@
...
+
+class ConstableClient : CliktCommand() {
+    private val path by argument(name = "h2-path", help = "Path to database file")
+    private val isQuery by option("-q",
+            help = "Specify if SQL command is a query and an output is required").flag(default = false)
+    private val command by option("--h2-command", help = "SQL command to execute").required()
+
+    override fun run() {
+        try {
+            if (isQuery)
```
dbarashev: You can use the same `execute` for everything. Its boolean return value indicates if there is a result set (use `getResultSet` to get it)
https://docs.oracle.com/javase/7/docs/api/java/sql/Statement.html#execute(java.lang.String).

com.bardsoftware.constable/client/src/main/kotlin/com/bardsoftware/constable/client/H2Manager.kt
```diff
@@ -0,0 +1,42 @@
...
+    /** Executes any SQL command and provides no output. */
+    fun execute(sqlCommand: String) {
+        connection.createStatement().use { statement -> statement.execute(sqlCommand) }
+    }
+
+    /** Executes SQL query that should start with SELECT operator and outputs result. */
+    fun executeQuery(sqlQuery: String) {
+        connection.createStatement().use { statement ->
+            statement.executeQuery(sqlQuery)
+                    .use { rs -> printResultSet(rs, System.out) }
```
dbarashev: .use on the previous line (like at line 18) will look better

com.bardsoftware.constable/client/src/main/kotlin/com/bardsoftware/constable/client/ConstableClient.kt
dbarashev: That's what shall be actually printed in the help string. Because now the help string says "path to the database file" which unambiguously means it is just a file.

com.bardsoftware.constable/client/src/main/kotlin/com/bardsoftware/constable/client/ConstableClient.kt
```diff
@@ -0,0 +1,33 @@
...
+                H2Manager(path).use { it.executeQuery(command) }
+            else {
+                H2Manager(path).use { it.execute(command) }
+            }
+        } catch (e: SQLException) {
+            logger.error(e) { "Something went wrong. ${e.message}" }
+            println("Something went wrong. " + e.message)
```
dbarashev: you don't need to println hat is already logged

PR #1715, opened 2020-02-24
Holiday calendars: Russia 2013, 2014, 2020, 2021

rbulygin:
Made some unification with Russian law (2-8 January is holidays every year)
Some weekends was missed for 2015, 2018 (when holiday on ordinal weekend)
2019 was corrupted
2013, 2014, 2020, 2021 was added
Sort of unification in descriptions 
I made it independently starting from master branch and then found you had already fixed some issues. But I hope this will be useful. Thank you


dbarashev:
> (2-8 January is holidays every year)

I can' be that certain about this ;).

As far as I understand, 2021 holidays will be officially announced somewhere in the middle of 2020 (thus, government's decision about 2020 holidays was published in July 2019) and Consultant's page is just a mere guess. For instance, there is an ongoing discussion about December 31st and chances are that it will become an official holiday, probably instead of some other holiday.

Are years prior to 2019 really important now :) ? Retro-planning?

PR #1716, opened 2020-02-24
Get rid of OSXAdapter

### LINKED ISSUE #1712: Get rid of OSXAdapter
Can we get rid of the [OSXAdapter](https://github.com/bardsoftware/ganttproject/blob/master/ganttproject/src/net/sourceforge/ganttproject/OSXAdapter.java) class? 

[com.apple.eawt.Application.addApplicationListener](https://github.com/bardsoftware/ganttproject/blob/master/ganttproject/src/net/sourceforge/ganttproject/OSXAdapter.java#L110) is not working anymore (it was deprecated in previous JDK versions)  and the rest of the methods are now included in java.awt.Desktop as of JDK 9 ([macOS-specific features have been removed in JDK 9](https://docs.oracle.com/javase/9/migrate/toc.htm#JSMIG-GUID-97C1D0BB-D5D3-4CAD-B17D-03A87A0AAF3B))

If accepted, I can add a PR to this issue but first I would like to know if in master version you are able to show the menubar of GanttProject integrated in macOS system bar (not in my case, even after calling `System.setProperty("apple.laf.useScreenMenuBar", "true"); `

In fact, I think that the setupSystemProperties method from OSXAdapter should be preserved... not sure.

Moreover, the handleOpenFile from OSXAdapter is supposed to  handle the case when a file in the Finder is dropped onto the app but it is not working anymore (without any changes to the master version). To double check it, I've also downloaded the 2.8.10 version, tried to drag&drop a .gan file to it with no luck (nothing happens).

Finally, I'd like to know if this kind of questions are better posted to the forum in help.ganttproject.biz or here... 

Thanks!

juananpe:
Fixes #1712


dbarashev:
Can we also remove AppleJavaExtensions.jar from biz.ganttproject.app.libs plugin?

ganttproject/src/net/sourceforge/ganttproject/GanttProject.java
```diff
@@ -665,11 +663,7 @@ void doShow() {
       Class.forName("java.awt.desktop.AboutHandler");
```
dbarashev: Since we're targeting at Java 11, I suggest removing this line and the surrounding try/catch block too. It was needed to be able to run under both Java 8 and Java 11 (there were no AboutHandler in Java 8), but now it is obsolete.

juananpe:
I have removed AppleJavaExtensions.jar from biz.ganttproject.app.libs and the building works as expected in macOS 10.15.3 so yep, I have added that removal as a new commit to the PR.

ganttproject/src/net/sourceforge/ganttproject/GanttProject.java
juananpe: Yes, I tested GP in macOS Catalina after removing it and it works as expected. It's great to help cleaning unused/unneeded lines of code :)

image

PR #1723, opened 2020-03-03
Tkt 1610 change logo without restart

upuente001:
Change the logo without restart

### DIFF
--- a/ganttproject/src/net/sourceforge/ganttproject/ChartTabContentPanel.java
+++ b/ganttproject/src/net/sourceforge/ganttproject/ChartTabContentPanel.java
@@ -62,6 +62,9 @@ abstract class ChartTabContentPanel {
       }
     });
   }
+  void replaceImagePanel(Image image){
+    myImagePanel.setIcon(new ImageIcon(image.getScaledInstance(-1, 50, Image.SCALE_DEFAULT)));
+  }
 
   JComponent createContentComponent() {
     JPanel tabContentPanel = new JPanel(new BorderLayout());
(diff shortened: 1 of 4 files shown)


ganttproject/src/net/sourceforge/ganttproject/ChartTabContentPanel.java
```diff
@@ -62,6 +62,9 @@ public void windowOpened(WindowEvent windowEvent) {
       }
     });
   }
+  void replaceImagePanel(Image image){
+    myImagePanel.setIcon(new ImageIcon(image.getScaledInstance(-1, 50, Image.SCALE_DEFAULT)));
```
dbarashev: any reason why the height must be 50px? There are calculations in the code below which take into account DPI setting. They probably shall not be ignored?

ganttproject/src/net/sourceforge/ganttproject/UIFacadeImpl.java
```diff
@@ -247,6 +236,23 @@ public void changeValue(ChangeValueEvent event) {
...
     myLogoOption = new DefaultFileOption("ui.logo");
+    myLogoOption.addChangeValueListener(new ChangeValueListener(){
+      @Override
+      public void changeValue(ChangeValueEvent event) {
+
+        if (event.getOldValue()!=null && event.getOldValue().equals(event.getNewValue())){
+          Image newLogo =getLogo();
+
+
+          ((GanttTree2)myFallbackDelegate.getTaskTree()).
```
dbarashev: I don't think that we want to expose that many internal fields and do all these casts to concrete classes and call chains. 
What about exposing LogoOption from UIFacade instead? Class `ChartTabContentPanel` has access to `UIFacade` instance, it will be able to get the logo option and add a listener.

PR #1740, opened 2020-05-27
Fixes #1596

### LINKED ISSUE #1596: Sort resources by name
It would be nice to sort resources in the resource chart by name.

Original request: https://help.ganttproject.biz/t/sort-resources-by-name/979

oihanealbizuri:
Fixes #1596


ganttproject/src/net/sourceforge/ganttproject/ResourceTreeTable.java
```diff
@@ -123,12 +131,93 @@ protected Chart getChart() {
...
   @Override
   protected void doInit() {
     super.doInit();
     myResourceTreeModel.updateResources();
+    getTable().getModel().addTableModelListener(new ModelListener());
     getVerticalScrollBar().addAdjustmentListener(new VscrollAdjustmentListener(myUiFacade.getResourceChart(), false));
+    getTableHeader().addMouseListener(new MouseAdapter() {
+      @Override
+      public void mouseClicked(MouseEvent mouseEvent) {
+        int index = getTable().columnAtPoint(mouseEvent.getPoint());
```
dbarashev, line 159: I understand this was copied from `GanttTreeTable`, right? Can you please refactor it somehow, so that the logic at lines 159-173 was common to both classes? One option is to extract this code to a method in `GPTreeTableBase` and call it from both listeners in the extended classes. Another option is to move the listener itself and this code to `GPTreeTableBase` and call an abstract method once execution passes all these checks -- and let abstract method implementations do all the remaining work.

ganttproject/src/net/sourceforge/ganttproject/ResourceTreeTable.java
```diff
@@ -123,12 +131,93 @@ protected Chart getChart() {
...
+          }
+        });
+      }
+    });
+  }
+
+  private class ModelListener implements TableModelListener {
+    @Override
+    public void tableChanged(TableModelEvent e) {
+      getUiFacade().getGanttChart().reset();
```
dbarashev, line 219: copy'n'paste failure, we're not in gantt chart here. Please refactor this too, so that there was common `ModelListener` in the base class.

ganttproject/src/net/sourceforge/ganttproject/ResourceDefaultColumn.java
```diff
@@ -53,6 +55,14 @@ private ResourceDefaultColumn(ColumnList.Column delegate, Class<?> valueClass, S
     myValueClass = valueClass;
   }
 
+  public Comparator<?> getSortComparator() {
```
dbarashev, line 58: Is it used anywhere besides the sorting code itself? Can we just keep a local comparator instance in that code, without storing the reference here?

ganttproject/src/net/sourceforge/ganttproject/ResourceTreeTable.java
```diff
@@ -85,18 +89,22 @@ private void setEditor(ColumnImpl column) {
     myResourceTreeModel.setSelectionModel(getTreeSelectionModel());
   }
 
+  private UIFacade getUiFacade() {
```
dbarashev, line 92: Please pull it up to `GPTreeTableBase` and make protected

ganttproject/src/net/sourceforge/ganttproject/ResourceTreeTable.java
```diff
@@ -123,12 +131,93 @@ protected Chart getChart() {
     return myUiFacade.getResourceChart();
   }
 
+  private static class AscendingNameComparator implements Comparator<HumanResource> {
```
dbarashev, line 134: Maybe use just one class? There is `reversed` method in any `Comparator` which makes, well, reversed comparator.

ganttproject/src/net/sourceforge/ganttproject/ResourceTreeTable.java
```diff
@@ -123,12 +131,93 @@ protected Chart getChart() {
...
+              if (column.getSort() == SortOrder.ASCENDING) {
+                column.setSort(SortOrder.DESCENDING);
+                resourceColumn.setSortComparator(new DescendingNameComparator());
+              } else {
+                column.setSort(SortOrder.ASCENDING);
+                resourceColumn.setSortComparator(new AscendingNameComparator());
+              }
+
+              ArrayList<HumanResource> sorted = new ArrayList<>();
+              getProject().getHumanResourceManager().getResources().forEach( h -> {
```
dbarashev, line 191: Why are you doing this one-by-one? If you want a copy, there is an `ArrayList` constructor which accepts a `Collection`

ganttproject/src/net/sourceforge/ganttproject/ResourceTreeTable.java
```diff
@@ -123,12 +131,93 @@ protected Chart getChart() {
...
+                sorted.add(h);
+              });
+              sorted.sort((Comparator<HumanResource>) resourceColumn.getSortComparator());
+
+              for(int j = sorted.size()-1; j>=0; j--){
```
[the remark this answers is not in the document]
dbarashev: You can add the parameter, if needed. Overloading may work too.

>  I can't possibly clear the resources and add them again because the nodes get lost 

Do you know exactly what's wrong with clearing (I mean, did you really observe anything wrong) ?

oihanealbizuri:
Thank you for your guidance!

PR #1743, opened 2020-06-01
GPCcsvExport test solved

Anaitz98:
the CSV export tests where outdated. The error was caused due to the lack of translation of the term "id". So i decided to return the original string in those cases, to avoid more possible errors.


ganttproject/src/biz/ganttproject/impex/csv/GanttCSVExport.java
```diff
@@ -153,7 +153,11 @@ public void save(SpreadsheetWriter writer) throws IOException {
   }
 
   private String i18n(String key) {
-    return GanttLanguage.getInstance().getText(key);
+    String result = GanttLanguage.getInstance().getText(key);
```
dbarashev, line 156: I suggest using `InternationalizationKt.getRootLocalizer().formatText(key)` instead. It returns the key if there's no translation.

ganttproject/src/biz/ganttproject/impex/csv/GanttCSVExport.java
```diff
@@ -230,6 +234,7 @@ private void writeTasks(SpreadsheetWriter writer) throws IOException {
   }
 
   private List<CustomPropertyDefinition> writeResourceHeaders(SpreadsheetWriter writer) throws IOException {
+    List<CustomPropertyDefinition> customFieldDefs = myHumanResourceCustomPropertyManager.getDefinitions();
```
dbarashev, line 237: This list is unused until line 253, why moving it here?

Anaitz98:
Moving the list was just for both methods ( writeResourceHeaders, writeTaskHeaders) to be more similar, but you're right it makes no sense to move it.

PR #1750, opened 2020-06-09
Added additional tests to test classes that have tight coupling with GUI

larsVanRoy:
Added a small test suite of classes that have high coupling with GUI instances. These were tested by overriding the classes that require user input in the GUI. A few additional scenario/regression tests were added for the same classes. Finally, refactoring was performed on the load class, as the task instance, was found counter intuitive (as it was mostly equal to a null pointer, and therefore not useable).


dbarashev:
Oh My God.
I am really sorry, but I will not review a pull request which touches 3k+ files and changes 800k+ lines of code. I think most of them are generated report, class files or whatever, but I really don't feel brave enough even to start investigating. 
 
If you believe that this work is really important, may I ask you to refactor the pull request  so that it became smaller? Pull requests are affordable when they change at most 1000 lines of code.

Thanks for your work!

LDVSOFT:
I am sorry to pass by, but it looks to me like it's the latest commit 09a2f61 and the first commit 
4167c9b that commit all those hundreds of thousands of lines. Commiting test & coverage reports is not considered a good practice, like commiting any other build artifact, especially very large ones. Clearing the branch (note that adding a commit that removes the reports would not clear the history, git would still store those) of reports while keeping test changes should make everything nice.

dbarashev:
No response. Closing this.

Issue #1295, opened 2016-08-04
Autosave files not deleted after program close


EdvinX:
I searched for this problem and found issue #656 "Please delete autosaves"
Unfortunaltely when I look to my TMP folder there are hundreds of: _ganttproject_autosaveXXXXXXXX.gan files.
Just last one _GanttProject_ps_XXXXX.gan is deleted after program close.

I'm using: GanttProject 2.7.2 Ostrava (build 1954)

dbarashev:
This is by design. If you start GanttProject again and close it immediately all autosave files will be deleted.

EdvinX:
I've just made a check by opening Ganttproject and closing it right after (without opening any project) and unfortunately about 300 of _ganttproject_autosaveXXXXXXXX.gan files remaines in TMP folder.

dbarashev:
What OS are you using?

dbarashev:
By the way, try waiting a little bit after opening GanttProject (a few minutes). The matter is that every GanttProject start collects autosave files created by previous launches. it is a background process which starts with a delay and needs to be completed.

EdvinX:
I'm using Win7.
I did as you suggested and closed GanttProject after about 40minutes, but all autosave files are still there.

PR #1752, opened 2020-06-17
Fixes #1508 and #1295 autosave files

### LINKED ISSUE #1508: Move autosave-files to a reboot-save location on Linux
Hi there,
when ganttproject crashes it is possible to recover unsaved changes due to the _ganttproject_autosaveXXX.gan files below /tmp . This is only usefull as long as the OS doesn't crash/reboot/runs out of battery.

When Linux starts, it will delete all files and folders below /tmp which includes ganttproject autosave-files. See [here](http://www.pathname.com/fhs/pub/fhs-2.3.html#TMPTEMPORARYFILES)

Probably /var/tmp would be a better choice, since files in this directory are *not* cleaned up after a reboot. See [here](http://www.pathname.com/fhs/pub/fhs-2.3.html#VARTMPTEMPORARYFILESPRESERVEDBETWEE).

On a side note (probably a new/different issue), it would be helpful, to check for autosave-files which are newer than the current one, when ganttproject opens a an existing project. I found it non-obvious to search for autosave-files after my system crashed the other day.

Thanks for the good work!

Anaitz98:
trying to solve the issue #1508 i  found that sometimes last accesed time is not changing so i  changed a little bit how it selects the files to delete.
modifiedTime

### DIFF
--- a/ganttproject/src/net/sourceforge/ganttproject/document/DocumentCreator.java
+++ b/ganttproject/src/net/sourceforge/ganttproject/document/DocumentCreator.java
@@ -12,6 +12,7 @@ import biz.ganttproject.core.table.ColumnList;
 import biz.ganttproject.core.time.CalendarFactory;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
+import com.jgoodies.common.base.SystemUtils;
 import net.sourceforge.ganttproject.GPLogger;
 import net.sourceforge.ganttproject.GanttOptions;
 import net.sourceforge.ganttproject.IGanttProject;
@@ -258,11 +259,13 @@ public class DocumentCreator implements DocumentManager {
 
   private static File getTempDir() {
     File tempDir;
-    if (System.getProperty("os.name").toLowerCase().equals("linux")){
+    if (SystemUtils.IS_OS_LINUX ){
       tempDir = new File("/var/tmp");
-    }else {
-      tempDir = new File(System.getProperty("java.io.tmpdir"));
+      if (tempDir.exists() && tempDir.isDirectory() && tempDir.canWrite()) {
+        return tempDir;
+      }
     }
+    tempDir = new File(System.getProperty("java.io.tmpdir"));
     if (tempDir.exists() && tempDir.isDirectory() && tempDir.canWrite()) {
       return tempDir;
     }
--- a/ganttproject/src/net/sourceforge/ganttproject/document/DocumentCreator.java
+++ b/ganttproject/src/net/sourceforge/ganttproject/document/DocumentCreator.java
@@ -207,7 +207,9 @@ public class DocumentCreator implements DocumentManager {
       GPLogger.log("Options file:" + optionsFile.getAbsolutePath());
       BasicFileAttributes attrs = Files.readAttributes(optionsFile.toPath(), BasicFileAttributes.class);
       FileTime accessTime = attrs.lastAccessTime();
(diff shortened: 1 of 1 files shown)


ganttproject/src/net/sourceforge/ganttproject/document/DocumentCreator.java
```diff
@@ -255,7 +257,12 @@ private FileSystem getAutosaveZipFs() {
   }
 
   private static File getTempDir() {
-    File tempDir = new File(System.getProperty("java.io.tmpdir"));
+    File tempDir;
+    if (System.getProperty("os.name").toLowerCase().equals("linux")){
+      tempDir = new File("/var/tmp");
```
dbarashev, line 262: I understand the need, however, we must be careful when working with paths which come not from Java. `/var/tmp` may not exist, it may exist but be not writable, we may or may not  have permissions to create it if it does not exist. Linux sysadmins may want to use some other directory instead, which is not possible when the value is hardcoded. 

Also, I doubt that `os.name` is always exactly (lowercased) `linux` when running on Linux.
  
Perhaps, it makes more sense to check if such directory exists and is writable and use it in case of success, otherwise just use the default temporary file directory.

ganttproject/src/net/sourceforge/ganttproject/document/DocumentCreator.java
```diff
@@ -207,7 +207,9 @@ public static Runnable createAutosaveCleanup() {
       GPLogger.log("Options file:" + optionsFile.getAbsolutePath());
       BasicFileAttributes attrs = Files.readAttributes(optionsFile.toPath(), BasicFileAttributes.class);
       FileTime accessTime = attrs.lastAccessTime();
-      cutoff = Math.min(accessTime.toMillis(), now);
+      FileTime modifyTime = attrs.lastModifiedTime();
+      long lastFileTime = Math.max(accessTime.toMillis(),modifyTime.toMillis());
```
dbarashev, line 211: please use spaces to separate function arguments.
```suggestion
      long lastFileTime = Math.max(accessTime.toMillis(), modifyTime.toMillis());
```

ganttproject/src/net/sourceforge/ganttproject/document/DocumentCreator.java
```diff
@@ -191,7 +191,7 @@ public Document newDocument(String path) throws IOException {
 
   @Override
   public Document newAutosaveDocument() throws IOException {
-    File tempFile = File.createTempFile("_ganttproject_autosave", ".gan");
+    File tempFile = File.createTempFile("_ganttproject_autosave", ".gan",getTempDir());
```
dbarashev, line 194:
```suggestion
    File tempFile = File.createTempFile("_ganttproject_autosave", ".gan", getTempDir());
```

ganttproject/src/net/sourceforge/ganttproject/document/DocumentCreator.java
Anaitz98: You mean not checking if OS is linux and just check if /var/tmp exists and its writable?

Or doing it once checked if OS its linux or not, mabe using SystemUtils.IS_OS_LINUX instead of os.name

dbarashev: I though about not checking, but perhaps SystemUtils + /var/tmp exists+is writable test is better.

PR #1781, opened 2020-09-26
SVG module

DmKazakov:
No changes to the code were made


dbarashev:
IMO "biz.ganttproject.mxgraph" is better name.

dbarashev (approved):

PR #1785, opened 2020-09-28
Removed getLoads() and getDaysOff() from LoadDistribution

DmKazakov:


dbarashev:
It seems that `getSeparatedTaskLoads` and `getResource` methods in `LoadDistribution` are also not used (except for a test).
However, it is probably not a good idea to stop testing pretty complext logic. Can you please update the tests so that they tested our new `calcLoadDistribution` function?

dbarashev (approved):
Great, thanks!

PR #1788, opened 2020-09-30
ResourcesColumnSceneBuilder

DmKazakov:


biz.ganttproject.core/src/biz/ganttproject/core/chart/scene/ResourcesColumnSceneBuilder.kt
```diff
@@ -0,0 +1,55 @@
...
+You should have received a copy of the GNU General Public License
+along with GanttProject.  If not, see <http://www.gnu.org/licenses/>.
+*/
+package biz.ganttproject.core.chart.scene
+
+import biz.ganttproject.core.chart.canvas.Canvas
+import biz.ganttproject.core.chart.canvas.TextMetrics
+import java.awt.Color
+
+private const val OFFSET = 5
```
dbarashev, line 25: This doesn't look as an offset of any sort, no? Just a default width of the empty list, divided by 2. I mean it is used only for this purpose.

biz.ganttproject.core/src/biz/ganttproject/core/chart/scene/ResourcesColumnSceneBuilder.kt
```diff
@@ -0,0 +1,55 @@
...
+private val ODD_ROW_COLOR = Color(0.933f, 0.933f, 0.933f)
```
dbarashev, line 26: Maybe replace with odd/even styles and assign colors using stylesheets?

biz.ganttproject.core/src/biz/ganttproject/core/chart/scene/ResourcesColumnSceneBuilder.kt
DmKazakov: It is meant to be value of left and right border offset

DmKazakov: Oh, I forgot to use this instead of the number when creating text

dbarashev (approved):

PR #1790, opened 2020-10-12
Don't use the same formatting for texts from the same line

DmKazakov:


biz.ganttproject.core/src/biz/ganttproject/core/chart/render/AbstractTextPainter.java
```diff
@@ -57,27 +57,53 @@ public void paint(TextGroup textGroup) {
...
+    } else if (line.size() == 1) {
+      paintWithMinLabel(line, leftX, bottomY, styles);
+    } else {
+      List<Text> middle = line.subList(1, line.size() - 1);
+      OptionalInt minLabel = paintWithMinLabel(middle, leftX, bottomY, styles);
+
+      Text first = line.get(0);
+      paintBorderLabel(first, minLabel, leftX, bottomY, styles);
+      Text last = line.get(line.size() - 1);
+      if (first != last) {
```
dbarashev, line 74: Did you really mean to use `!=` ? It is Java, not Kotlin. `!=` != `equals` in Java. And in any case, is it possible that first and last texts are equals to each other in any sense if there are more than 1 text in the line?

DmKazakov: Yes, I meant to compare references. Probably I wanted to reduce number of branches and forgot to delete it when it didn't work out

dbarashev (approved):

PR #1813, opened 2021-01-25
Painting multiple columns in TableSceneBuilder

DmKazakov:


biz.ganttproject.core/src/main/java/biz/ganttproject/core/chart/scene/gantt/TableSceneBuilder.kt
```diff
@@ -19,40 +19,104 @@ along with GanttProject.  If not, see <http://www.gnu.org/licenses/>.
...
+import biz.ganttproject.core.chart.canvas.TextMetrics
+import java.lang.Integer.max
 
 class TableSceneBuilder(private val config: Config) {
-  fun build(rows: List<Row>, canvas: Canvas = Canvas()): Canvas {
-    val dimensions = calculateDimensions(rows)
+  private lateinit var table: Table
+  private lateinit var canvas: Canvas
+  private lateinit var colsWidth: Map<Table.Column, Int>
+  private lateinit var dimensions: Dimension
```
DmKazakov, line 29: Didn't like passing this 4 parameters to the paint methods, so ended up turning them into fields

biz.ganttproject.core/src/main/java/biz/ganttproject/core/chart/scene/gantt/TableSceneBuilder.kt
```diff
@@ -19,40 +19,104 @@ along with GanttProject.  If not, see <http://www.gnu.org/licenses/>.
...
+          val indent = if (isFirst) it.indent else 0
+          indent + (it.values[col]?.run { config.textMetrics.getTextLength(this) } ?: 0)
+        }.maxOrNull() ?: 0
+        max(colNameWidth, colContentWidth)
+      }
+    }
+    return widths
+  }
+
+  private fun calculateDimensions(table: Table, colsWidth: Map<Table.Column, Int>): Dimension {
```
dbarashev, line 71: `colsWidth` is passed here in the arguments while in `paintHeader` it is used as a membe variable. I think it is better to stick to a single way of using them.

biz.ganttproject.core/src/main/java/biz/ganttproject/core/chart/scene/gantt/TableSceneBuilder.kt
dbarashev: This means that we need to be careful with using the instance of this builder and do not allow for concurrent calls to `build`. In particular, it will make sense to construct the instance every time in `TaskTableSceneBuilder::build` instead of using a single instance as it is done now.

biz.ganttproject.core/src/main/java/biz/ganttproject/core/chart/scene/gantt/TaskTableSceneBuilder.kt
```diff
@@ -19,31 +19,45 @@ along with GanttProject.  If not, see <http://www.gnu.org/licenses/>.
...
+import biz.ganttproject.core.chart.scene.gantt.TableSceneBuilder.Table.*
 import biz.ganttproject.core.chart.canvas.Canvas
 import biz.ganttproject.core.chart.canvas.TextMetrics
 
 class TaskTableSceneBuilder(
   private val input: InputApi
 ) {
-  private val tableSceneBuilder = TableSceneBuilder(Config(input.rowHeight, input.horizontalOffset))
+  private val tableSceneBuilder = TableSceneBuilder(Config(input.rowHeight, input.horizontalOffset, input.textMetrics))
+  private val cols = listOf(Column("Name"), Column("Begin date"), Column("End date"), Column("Cost"))
```
dbarashev, line 30: Is this list fixed just for testing? In reality it needs to be configurable, of course.

biz.ganttproject.core/src/main/java/biz/ganttproject/core/chart/scene/gantt/TaskTableSceneBuilder.kt
```diff
@@ -19,31 +19,45 @@ along with GanttProject.  If not, see <http://www.gnu.org/licenses/>.
...
+          cols[0] to it.name, cols[1] to it.beginDate, cols[2] to it.endDate, cols[3] to it.cost
+        ),
         indent
-      )) + toRows(it.subtasks, indent + input.depthIndent)
+      )) + toRow(it.subtasks, indent + input.depthIndent)
     }
   }
 
-  class Task(val name: String, val subtasks: List<Task> = emptyList()) {
+  class Task(
```
dbarashev, line 52: We'll end up with one more interface for tasks this way... Not sure if it makes sense, because what matters for these  painting purposes are strings and probably their alignment, and it doesn't matter that a string is "name" or "cost". Maybe stay with the interface providing Column => String mapping ?

biz.ganttproject.core/src/main/java/biz/ganttproject/core/chart/scene/gantt/TableSceneBuilder.kt
DmKazakov: Then maybe move `table` to constructor arguments and create a `TableSceneBuilder`  for each `build` call?

biz.ganttproject.core/src/main/java/biz/ganttproject/core/chart/scene/gantt/TableSceneBuilder.kt
DmKazakov: It was done so that it's harder to mess up with `lateinit`, but apparently we will get rid of it

biz.ganttproject.core/src/main/java/biz/ganttproject/core/chart/scene/gantt/TableSceneBuilder.kt
dbarashev: Sounds reasonable.

dbarashev:
LGTM

PR #1815, opened 2021-02-03
Spreadsheet export/import bugfixes

dbarashev:
So, this started from fixing Excel export which created all cells as text strings. Excel supports a variety of data types: text, boolean, numeric (same as decimal, double precision) and with some tricks date, so it would be natural to use these types.

During the work it turned out that there were major issues with resource custom column import, and they were fixed in the same branch.


dbarashev:
Updated plugin: `ganttproject`

PR #1817, opened 2021-02-09
UnboundedViewportTaskScene

DmKazakov:


biz.ganttproject.mxgraph/src/main/kotlin/biz/ganttproject/mxgraph/SummaryTaskPainter.kt
```diff
@@ -0,0 +1,60 @@
...
+      mxPainterImpl.paintRectangle(
+        rectangle.leftX + left, rectangle.topY + top,
+        rectangle.width - (left + right), rectangle.height - (top + bottom),
+        mxStyle, rectangle.attributes
+      )
+
+      val notchWidth = rectangle.height - (top + bottom)
+      if (rectangle.hasStyle("task.summary.open")) {
+        mxPainterImpl.paintRectangle(
+          rectangle.leftX, rectangle.topY, notchWidth, rectangle.height,
```
dbarashev, line 48: How will it look like graphically?

ganttproject/src/main/java/net/sourceforge/ganttproject/chart/UnboundedViewportTaskSceneBuilder.kt
```diff
@@ -0,0 +1,184 @@
...
+        else -> null.also { logger?.invoke("Connector-activity mismatch") }
+      }
+    }
+
+    override fun getStyle(dependency: TaskDependency): String {
+      return if (dependency.dependant.hardness === Hardness.STRONG) "dependency.line.hard" else "dependency.line.rubber"
+    }
+
+    override fun getConnectors(task: Task): Iterable<TaskDependency> {
+      return task.dependants.map { TaskDependency(task, it) }
```
dbarashev, line 125: Is it a typo or as designed? The prototype code in Java calls `task.getDependencies()` (all deps, not only dependants).

ganttproject/src/main/java/net/sourceforge/ganttproject/chart/UnboundedViewportTaskSceneBuilder.kt
```diff
@@ -0,0 +1,184 @@
...
+
+    inner class Activity : BarChartActivity<Task> {
+      override fun getStart() = this@Task.start
+      override fun getEnd() = this@Task.end
+      override fun getOwner() = this@Task
+
+      override fun getDuration(): TimeDuration {
+        return TimeDurationImpl(
+          GPTimeUnitStack.DAY,
+          TimeUnit.DAYS.convert(start.time - end.time, TimeUnit.MILLISECONDS)
```
dbarashev, line 97: I don't think that we want to reinvent duration building. There is `TaskManager::createDuration` method and I think it makes sense to get duration from there, via `Task` interface.

ganttproject/src/main/java/net/sourceforge/ganttproject/chart/UnboundedViewportTaskSceneBuilder.kt
```diff
@@ -0,0 +1,184 @@
...
+  abstract class Task {
+    abstract val start: Date
+    abstract val end: Date
+    abstract val isMilestone: Boolean
+    abstract val subtasks: List<Task>
+    abstract val dependants: List<Dependant>
+
+    val activity = Activity()
+
+    fun flatten(result: MutableList<Task> = mutableListOf()): List<Task> {
```
dbarashev, line 83: this is basically pre-order DFS? There's `TaskContainmentHierarchyFacade#getTasksInDocumentOrder` which I think should be used.

biz.ganttproject.mxgraph/src/main/kotlin/biz/ganttproject/mxgraph/SummaryTaskPainter.kt
DmKazakov: image
I did it like this since mxGraph doesn't support right-angled triangles out of the box and custom shapes don't make their way to xml.

ganttproject/src/main/java/net/sourceforge/ganttproject/chart/UnboundedViewportTaskSceneBuilder.kt
DmKazakov: It's intentional, this way we go through all dependencies anyway. I tested it on HouseBuildingSample.gan and it works as expected. Looks like the original code does extra work here or I just haven't notice some special cases.

ganttproject/src/main/java/net/sourceforge/ganttproject/chart/UnboundedViewportTaskSceneBuilder.kt
DmKazakov: I think `Task` interface requires too much to implement it here.

PR #1862, opened 2021-04-01
Add a dependency on Open JavaFX (#2)

brodeuralexis:
This commit enables the use of OpenJDK 11+ and other JDKs when compiling and
executing the project.

This commit adds a dependency to Open JavaFX, the open-source
implementation of the JavaFX API by the OpenJDK organization.

### DIFF
--- a/ganttproject/build.gradle
+++ b/ganttproject/build.gradle
@@ -8,6 +8,7 @@ buildscript {
     }
     dependencies {
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
+        classpath 'org.openjfx:javafx-plugin:0.0.9'
     }
 }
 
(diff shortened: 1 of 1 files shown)


dbarashev:
Okay, at the first glance seems to be working and not breaking the builds, at least on Linux. Let's see how it goes.

brodeuralexis:
I'm still trying to understand why the macOS build fails.

It boggles my mind that `jpackage` does not seem to be present in the macOS distribution of OpenJDK 11.

dbarashev:
There's no jpackage at all in OpenJDK 11. Its history was long and full of adventures: https://openjdk.java.net/jeps/392
I build distribution packages using BellSoft Liberica Java 11 for compiling and Java 14 for running jpackage.

PR #2076, opened 2022-04-02
Insert new tasks into in-memory DB

npostnikova:


npostnikova:
Not all the task fields have been added yet. But it's better to learn now whether I'm doing something completely wrong :)

ganttproject/src/main/java/net/sourceforge/ganttproject/ProjectStateHolderEventListener.kt
```diff
@@ -19,24 +19,72 @@ along with GanttProject.  If not, see <http://www.gnu.org/licenses/>.
...
  */
-class ProjectStateHolderEventListener(private val databaseFactory: () -> ProjectDatabase) : Stub() {
-  private val lazyProjectDatabase = lazy { databaseFactory()  }
-  private val projectDatabase: ProjectDatabase by lazyProjectDatabase
-
-  override fun projectClosed() {
-    if (lazyProjectDatabase.isInitialized()) {
-      // ...
-      projectDatabase.shutdown()
+class ProjectStateHolderEventListener(private val projectDatabase: ProjectDatabase) : TaskListener, Stub() {
```
npostnikova, line 31: Seems like there're a lot of listeners yet to add, so it's better to create a separate class for each interface, passing the same db, right?

ganttproject/src/main/java/net/sourceforge/ganttproject/storage/SqlProjectDatabaseImpl.kt
```diff
@@ -0,0 +1,99 @@
...
+    fun createInMemoryDatabase(): ProjectDatabase {
+      val dataSource = JdbcDataSource()
+      dataSource.setURL(H2_IN_MEMORY_URL)
+      return SqlProjectDatabaseImpl(dataSource)
+    }
+  }
+
+  private fun <T> withExecutor(
+    errorMessage: String = "Failed to execute query",
+    body: (executor: DSLContext) -> T
```
dbarashev, line 47: also, minor nitpicking: the name "executor" in JVM apps is usually taken by ExecutorService and its friends, and GP is no exception. jOOQ named this object "DSLContext", so perhaps we can call it "dsl" too

ganttproject/src/test/java/net/sourceforge/ganttproject/storage/ProjectDatabaseTest.kt
```diff
@@ -0,0 +1,115 @@
...
+        .selectFrom(TASK)
+        .fetch()
+    assert(tasks.isEmpty())
+  }
+
+  @Test
+  fun `test insert task`() {
+    projectDatabase.init()
+
+    val mockTask: GanttTask = niceMock(GanttTask::class.java)
```
dbarashev, line 68: in fact, you can use the real tasks instead of mocks. There is a useful `TestSetupHelper` class which can create an instance of TaskManager for you

dbarashev:
Good job, overall!

ganttproject/src/main/resources/sql/init-project-database.sql
```diff
@@ -0,0 +1,5 @@
+create table if not exists Task (
+    id      integer     not null    primary key,
+    name    varchar     not null,
```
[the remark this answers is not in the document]
npostnikova: I've missed the argument, sorry. It seems to be [2147483647](https://groups.google.com/g/h2-database/c/2TWZMSNLGqg?pli=1). Of course, it's bad to rely on this non-standardized value.
What length sounds reasonable?

ganttproject/src/test/java/net/sourceforge/ganttproject/storage/ProjectDatabaseTest.kt
npostnikova: That's cool!
Seems like the test file should be moved to `ganttproject-tester`, right?

ganttproject/src/main/java/net/sourceforge/ganttproject/ProjectStateHolderEventListener.kt
npostnikova: Should we modify it later if required?

dbarashev (approved):
LGTM

PR #2081, opened 2022-04-07
Create and insert queries for Task & TaskDependency

npostnikova:
* Add the rest of the fields to the Task table
* Create TaskDependency table

### DIFF
--- a/ganttproject/build.gradle
+++ b/ganttproject/build.gradle
@@ -79,6 +79,9 @@ jooq {
                     target {
                         packageName = 'biz.ganttproject.storage.db'
                     }
+                    generate {
+                        javaTimeTypes = false
+                    }
                 }
             }
         }
--- a/ganttproject/src/main/java/net/sourceforge/ganttproject/storage/ProjectDatabase.kt
+++ b/ganttproject/src/main/java/net/sourceforge/ganttproject/storage/ProjectDatabase.kt
@@ -19,7 +19,9 @@ along with GanttProject.  If not, see <http://www.gnu.org/licenses/>.
 
 package net.sourceforge.ganttproject.storage
 
+import net.sourceforge.ganttproject.CustomPropertyDefinition
 import net.sourceforge.ganttproject.task.Task
+import net.sourceforge.ganttproject.task.dependency.TaskDependency
 import kotlin.jvm.Throws
 
 open class ProjectDatabaseException(message: String): Exception(message)
@@ -33,6 +35,14 @@ interface ProjectDatabase {
   /** Insert the task. */
   @Throws(ProjectDatabaseException::class)
   fun insertTask(task: Task)
+
+  /** Insert the task dependency. */
+  @Throws(ProjectDatabaseException::class)
+  fun insertTaskDependency(taskDependency: TaskDependency)
+
+  /** Insert the custom property. */
+  @Throws(ProjectDatabaseException::class)
+  fun insertCustomProperty(task: Task, customProperty: CustomPropertyDefinition)
 
   /** Close connections and release the resources. */
   @Throws(ProjectDatabaseException::class)
--- a/ganttproject/src/main/java/net/sourceforge/ganttproject/storage/ProjectDatabase.kt
+++ b/ganttproject/src/main/java/net/sourceforge/ganttproject/storage/ProjectDatabase.kt
@@ -19,7 +19,6 @@ along with GanttProject.  If not, see <http://www.gnu.org/licenses/>.
 
 package net.sourceforge.ganttproject.storage
 
-import net.sourceforge.ganttproject.CustomPropertyDefinition
 import net.sourceforge.ganttproject.task.Task
 import net.sourceforge.ganttproject.task.dependency.TaskDependency
 import kotlin.jvm.Throws
@@ -39,10 +38,6 @@ interface ProjectDatabase {
   /** Insert the task dependency. */
   @Throws(ProjectDatabaseException::class)
   fun insertTaskDependency(taskDependency: TaskDependency)
-
-  /** Insert the custom property. */
-  @Throws(ProjectDatabaseException::class)
-  fun insertCustomProperty(task: Task, customProperty: CustomPropertyDefinition)
 
   /** Close connections and release the resources. */
   @Throws(ProjectDatabaseException::class)
--- a/ganttproject/src/main/java/net/sourceforge/ganttproject/storage/SqlProjectDatabaseImpl.kt
+++ b/ganttproject/src/main/java/net/sourceforge/ganttproject/storage/SqlProjectDatabaseImpl.kt
@@ -19,20 +19,28 @@ along with GanttProject.  If not, see <http://www.gnu.org/licenses/>.
 
 package net.sourceforge.ganttproject.storage
 
+import biz.ganttproject.core.time.GanttCalendar
 import biz.ganttproject.storage.db.Tables.*
+import com.google.common.base.Charsets
+import net.sourceforge.ganttproject.CustomPropertyDefinition
 import net.sourceforge.ganttproject.GPLogger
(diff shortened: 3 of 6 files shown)


ganttproject/build.gradle
```diff
@@ -79,6 +79,9 @@ jooq {
                     target {
                         packageName = 'biz.ganttproject.storage.db'
                     }
+                    generate {
+                        javaTimeTypes = false
```
dbarashev, line 83: Can you please add a comment explaining this trick?

ganttproject/src/main/java/net/sourceforge/ganttproject/storage/SqlProjectDatabaseImpl.kt
```diff
@@ -20,19 +20,23 @@ along with GanttProject.  If not, see <http://www.gnu.org/licenses/>.
 package net.sourceforge.ganttproject.storage
 
 import biz.ganttproject.storage.db.Tables.*
+import com.google.common.base.Charsets
```
dbarashev, line 23: I suggest using Charsets object from Kotlin stdlib (kotlin.text.Charsets)

ganttproject/src/main/java/net/sourceforge/ganttproject/storage/SqlProjectDatabaseImpl.kt
```diff
@@ -69,17 +73,81 @@ class SqlProjectDatabaseImpl(private val dataSource: DataSource): ProjectDatabas
...
+      .column(TASK.THIRD_DATE_CONSTRAINT, INTEGER.null_())
+      .column(TASK.PRIORITY, VARCHAR.null_())
+      .column(TASK.WEB_LINK, VARCHAR.null_())
+      .column(TASK.COST_MANUAL_VALUE, VARCHAR.null_())
+      .column(TASK.IS_COST_CALCULATED, BOOLEAN.null_())
+      .column(TASK.NOTES, VARCHAR.null_())
       .constraints(primaryKey(TASK.ID))
       .execute()
+
+    dsl.createTable(TASKDEPENDENCY)
```
dbarashev, line 92: Any sense to use same schema defined in `init-project-database.sql` instead of duplicating it?

ganttproject/build.gradle
npostnikova: Added a comment
I'm not sure that the flag is necessary but the date/time conversion seems to be nicer this way as `java.sql.Timestamp` can be constructed using `timeInMillis`. Without it, jooq required using `java.time.LocalDateTime`

ganttproject/src/main/java/net/sourceforge/ganttproject/storage/SqlProjectDatabaseImpl.kt
```diff
@@ -63,23 +66,70 @@ class SqlProjectDatabaseImpl(private val dataSource: DataSource): ProjectDatabas
...
+    val webLink = task.webLink.let { link ->
+      if (!link.isNullOrBlank() && link != "http://") URLEncoder.encode(link, Charsets.UTF_8.name()) else null
+    }
+    var costManualValue: String? = null
+    var isCostCalculated: Boolean? = null
+    if (!(task.cost.isCalculated && task.cost.manualValue == BigDecimal.ZERO)) {
+      costManualValue = task.cost.manualValue.toPlainString()
+      isCostCalculated = task.cost.isCalculated
+    }
+    val notes = task.notes?.let { notes -> notes.replace("\\r\\n", "\\n").ifBlank { null } }
```
dbarashev, line 101: This was basically copied from TaskSaver, right? Can you please add a TODO to use common serialization code in the future?

ganttproject/src/main/resources/resources/sql/init-project-database.sql
```diff
@@ -0,0 +1,31 @@
...
+    name                    varchar         not null,
+    color                   varchar         null,
+    shape                   varchar         null,
+    is_milestone            boolean         not null,
+    is_project_task         boolean         not null,
+    start_date              timestamp(0)    not null,
+    duration                integer         not null,
+    completion              integer         null,
+    earliest_start_date     timestamp(0)    null,
+    third_date_constraint   integer         null,
```
dbarashev, line 12: this field is useless (it is always "1" when the earliest start date is set, and takes no other values).

dbarashev (approved):
There are a few relatively minor comments, otherwise looks good. Feel free to merge after resolving the comments.

ganttproject/src/main/java/net/sourceforge/ganttproject/storage/SqlProjectDatabaseImpl.kt
```diff
@@ -63,23 +66,70 @@ class SqlProjectDatabaseImpl(private val dataSource: DataSource): ProjectDatabas
...
+    try {
+      val queries = String(scriptStream.readAllBytes(), Charsets.UTF_8)
+
+      dataSource.connection.use { connection ->
+        connection.createStatement().execute(queries)
+      }
+    } catch (e: Exception) {
+      val message = "Failed to init the database {}"
+      LOG.error(message, e)
+      throw ProjectDatabaseException(message)
```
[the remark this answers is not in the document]
npostnikova: Thanks for the comments! 
Though, I'm not sure that I understood this one correctly. Could you please check the last commit?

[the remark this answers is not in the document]
dbarashev: the message is "Failed to init the database {}". When logged at line 84, with the exception instance, SLF4J handles the exception instance properly and we see the stack trace or something like that. However, when passed to the `ProjectDatabaseException` constructor, it is still literally ""Failed to init the database {}".

Besides, we're going to log the error twice, once here and once at the place where we handle ProjectDatabaseException. I suggest creating a `ProjectdatabaseException` constructor with `cause: Exception` argument, pass the exception which we catch here as the cause and remove logging here.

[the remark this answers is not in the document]
dbarashev: I see that you pass the cause in `init` method, but in `withDSL` it still works as described above

ganttproject/src/main/java/net/sourceforge/ganttproject/storage/ProjectDatabase.kt
```diff
@@ -20,9 +20,12 @@ along with GanttProject.  If not, see <http://www.gnu.org/licenses/>.
 package net.sourceforge.ganttproject.storage
 
 import net.sourceforge.ganttproject.task.Task
+import net.sourceforge.ganttproject.task.dependency.TaskDependency
 import kotlin.jvm.Throws
 
-open class ProjectDatabaseException(message: String): Exception(message)
+open class ProjectDatabaseException(message: String): Exception(message) {
+  constructor(message: String, cause: Exception): this("$message. Caused by ${cause.message}")
```
dbarashev, line 27: There is a super-constructor with the cause argument. It creates a chain of exceptions and prints the stack traces of all exceptions in the chain.

ganttproject/src/main/resources/resources/sql/init-project-database.sql
```diff
@@ -0,0 +1,30 @@
...
+    shape                   varchar         null,
+    is_milestone            boolean         not null,
+    is_project_task         boolean         not null,
+    start_date              timestamp(0)    not null,
+    duration                integer         not null,
+    completion              integer         null,
+    earliest_start_date     timestamp(0)    null,
+    priority                varchar         null,
+    web_link                varchar         null,
+    cost_manual_value       varchar         null,
```
dbarashev, line 14: I believe that numeric is better type for the cost values.

dbarashev (approved):
LGTM

PR #2087, opened 2022-04-14
Tkt 2083 | Extract common task serialization code

npostnikova:
* Convert TaskSaver to Kotlin
* Make TaskSaver and ProjectDatabase share the code responsible for conversion of task fields


ganttproject/src/main/java/net/sourceforge/ganttproject/task/TaskSerializationUtils.kt
```diff
@@ -0,0 +1,42 @@
...
+  return null
+}
+
+// XML CDATA section adds extra line separator on Windows.
+// See https://bugs.openjdk.java.net/browse/JDK-8133452.
+fun Task.externalizedNotes(): String? = notes?.replace("\\r\\n", "\\n")?.ifBlank { null }
+
+fun TaskImpl.externalizedColor(): String? = if (colorDefined()) ColorConvertion.getColor(color) else null
+
+fun Task.externalizedStartDate(): Timestamp = Timestamp(start.timeInMillis)
```
dbarashev, line 40: I suggest moving these dates conversions to SQL specific code, because it is not useful for XML serialization

ganttproject/src/main/java/net/sourceforge/ganttproject/task/TaskSerializationUtils.kt
```diff
@@ -0,0 +1,37 @@
...
+along with GanttProject.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+package net.sourceforge.ganttproject.task
+
+import net.sourceforge.ganttproject.util.ColorConvertion
+import java.net.URLEncoder
+
+
+fun Task.externalizedWebLink(): String? {
```
dbarashev, line 26: move the remaining methods into TaskSaver.kt?

dbarashev (approved):
LGTM

PR #2090, opened 2022-04-18
Added Task::getUid method

dbarashev:
... and returned back lots of tests :)

### DIFF
--- a/ganttproject-tester/test/biz/ganttproject/core/calendar/WeekendCalendarImplTest.java
+++ b/ganttproject-tester/test/biz/ganttproject/core/calendar/WeekendCalendarImplTest.java
@@ -18,28 +18,27 @@ along with GanttProject.  If not, see <http://www.gnu.org/licenses/>.
 */
 package biz.ganttproject.core.calendar;
 
+import biz.ganttproject.core.calendar.GPCalendar.DayMask;
+import biz.ganttproject.core.calendar.GPCalendar.DayType;
+import biz.ganttproject.core.time.CalendarFactory;
+import com.google.common.collect.Collections2;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
+import junit.framework.TestCase;
+
 import java.text.DateFormat;
 import java.util.Calendar;
 import java.util.List;
 import java.util.Locale;
-
-import biz.ganttproject.core.calendar.GPCalendar.DayMask;
-import biz.ganttproject.core.calendar.GPCalendar.DayType;
-import biz.ganttproject.core.time.CalendarFactory;
-
-import com.google.common.base.Function;
-import com.google.common.collect.Collections2;
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Lists;
-
-import junit.framework.TestCase;
+import java.util.function.Function;
 
 /**
- * Tests for {@link WeekendsCalendarImpl} class.
+ * Tests the implementation of a calendar with weekends and holidays.
  *
  * @author dbarashev (Dmitry Barashev)
  */
 public class WeekendCalendarImplTest extends TestCase {
+  @Override
   protected void setUp() {
     new CalendarFactory() {
       {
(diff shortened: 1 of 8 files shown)


ganttproject/src/main/java/net/sourceforge/ganttproject/task/TaskManagerImpl.java
```diff
@@ -355,15 +355,18 @@ public TaskBuilder newTaskBuilder() {
...
-
         if (myPrototype != null) {
           myId = myPrototype.getTaskID();
         }
         if (myId == null || myTaskMap.getTask(myId) != null) {
           myId = getAndIncrementId();
         }
 
-        TaskImpl task = new GanttTask("", CalendarFactory.createGanttCalendar(), 1, TaskManagerImpl.this, myId);
+        var startDate = myStartDate != null
```
npostnikova, line 365: Should this part of the code be removed from `TaskManager.TaskBuilder.setupNewTask` then? The method is not used anywhere but here

ganttproject-tester/test/biz/ganttproject/core/calendar/WeekendCalendarImplTest.java
```diff
@@ -56,15 +56,19 @@ public DateFormat getShortDateFormat() {
...
       CalendarEvent.newEvent(CalendarFactory.createGanttCalendar(2014, 1, 14).getTime(), false, CalendarEvent.Type.NEUTRAL, "Feb 14", null),
       CalendarEvent.newEvent(CalendarFactory.createGanttCalendar(2014, 2, 8).getTime(), true, CalendarEvent.Type.HOLIDAY, "Mar 8", null),
       CalendarEvent.newEvent(CalendarFactory.createGanttCalendar(2014, 2, 8).getTime(), false, CalendarEvent.Type.WORKING_DAY, "Mar 8, 2014", null),
       CalendarEvent.newEvent(CalendarFactory.createGanttCalendar(2014, 2, 9).getTime(), false, CalendarEvent.Type.HOLIDAY, "Mar 9, 2014", null),
       CalendarEvent.newEvent(CalendarFactory.createGanttCalendar(2014, 3, 12).getTime(), true, CalendarEvent.Type.WORKING_DAY, "Apr 12", null),
       CalendarEvent.newEvent(CalendarFactory.createGanttCalendar(2014, 3, 12).getTime(), false, CalendarEvent.Type.HOLIDAY, "Apr 12, 2014", null)
-  );
+    );
+  }
+  private static List<CalendarEvent> TEST_EVENTS = createTestEvents();
```
npostnikova, line 71: Could you please explain the trick? :)
Why to create `TEST_EVENTS` and use it in the first method only?
Shouldn't `TEST_EVENTS_RECURRING_FIRST` be converted to method as well for consistency?

ganttproject-tester/test/biz/ganttproject/core/calendar/WeekendCalendarImplTest.java
```diff
@@ -40,7 +40,7 @@
  * @author dbarashev (Dmitry Barashev)
  */
 public class WeekendCalendarImplTest extends TestCase {
-  static {
+  protected void setUp() {
```
npostnikova, line 43: An explicit override would be nice :)

npostnikova (approved):
Looks good overall!

ganttproject-tester/test/biz/ganttproject/core/calendar/WeekendCalendarImplTest.java
dbarashev: Thanks, I replaced both static lists with the method calls.
The issue which I struggled with was that these tests were failing if I ran them all (like `gradle test`). However, with a filter (even a very simple `gradle test --tests biz*`) they were passing. 
I didn't get the reasons exactly, but I think that there were side effects of setting the static member `ourLocaleApi` from different tests. That's why I replaced static instances and static setup with instances created per each test.

ganttproject/src/main/java/net/sourceforge/ganttproject/task/TaskManagerImpl.java
dbarashev: done, thanks

ganttproject-tester/test/biz/ganttproject/core/calendar/WeekendCalendarImplTest.java
dbarashev: yup

PR #2091, opened 2022-04-19
Tkt 2084 | Use date data type for Task.startDate & Task.thirdDate

npostnikova:
* Replace timestamp with date
* Weirdly fix some issues with storing dates via executing a string representation of sql queries ([similar issue](https://github.com/h2database/h2database/issues/2573)). Well, I planned to do it further anyway, so maybe it's fine :)


ganttproject/src/main/java/net/sourceforge/ganttproject/storage/SqlProjectDatabaseImpl.kt
```diff
@@ -231,9 +233,7 @@ class SqlTaskUpdateBuilder(private val task: Task,
   }
 }
 
-private fun Task.externalizedStartDate(): Timestamp = Timestamp(start.timeInMillis)
-
-private fun Task.externalizedEarliestStartDate(): Timestamp? = third?.let { calendar -> Timestamp(calendar.timeInMillis)  }
+private fun GanttCalendar.externalizeDate(): LocalDate = LocalDate.parse(DateParser.getIsoDateNoHours(time))
```
dbarashev, line 236: maybe rename to `toLocalDate` (and even move to GanttCalendar) ?

dbarashev (approved):
LGTM! One naming suggestion.

Issue #2089, opened 2022-04-18
Introduce UUID identifier for tasks


dbarashev:
We need UUIDs to decrease the probability of conflicts in the real-time collaboration, e.g. when two users concurrently create a new task. We will distinguish tasks by UUIDs when sending them to the server and back, however, we'll keep the current ID as the task identifier in the desktop GanttProject. We will allow for the ID duplicates at merge time and will renumber tasks during merging.

Issue #2084, opened 2022-04-11
Update H2 database on task property changes


dbarashev:
Rough plan: pass a factory of SQL statements or jOOQ DSL intances to the task manager and generate the appropriate update statements in Task.MutatorImpl for every call of a property setter.

Besides the update itself, we need to keep the log of the update statements and later send it to the Cloud for syncing.

PR #2100, opened 2022-04-22
Tkt 2084 | Use task uid as primary key

npostnikova:


dbarashev:
Update #2084 #2089

dbarashev (approved):
LGTM

PR #2101, opened 2022-04-22
Add WebSocket skeleton on the server side

npostnikova:


npostnikova:
@dbarashev Am I doing something weird?

dbarashev:
Pls add a dependency on nanohttpd-websocket

PR #2102, opened 2022-04-23
Add naive xlog sending to WebSocketClient

npostnikova:


ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
```diff
@@ -880,4 +895,21 @@ public void refresh() {
     super.repaint();
   }
 
+  private Unit sendProjectStateLogs() {
```
npostnikova, line 898: It's not normal, is it? :)
Is it better to replace `this::sendProjectStateLogs` usage instead?

ganttproject/src/main/java/biz/ganttproject/storage/cloud/GPCloudHttpImpl.kt
```diff
@@ -262,18 +267,27 @@ class WebSocketClient {
   private val structureChangeListeners = mutableListOf<(Any) -> Unit>()
   private val lockStatusChangeListeners = mutableListOf<(ObjectNode) -> Unit>()
   private val contentChangeListeners = mutableListOf<(ObjectNode) -> Unit>()
+  private val xlogReceivedListeners = mutableListOf<(ObjectNode) -> Unit>()
   private var listeningDocument: GPCloudDocument? = null
 
-  fun start() {
+  private fun getWebSocketUrl(): String {
+    return if (isColloboqueLocalTest()) {
```
dbarashev, line 274: more Kotlin'ish style: 

```suggestion
  private fun getWebSocketUrl() = if (isColloboqueLocalTest()) {
```

ganttproject/src/main/java/biz/ganttproject/storage/cloud/GPCloudHttpImpl.kt
```diff
@@ -262,18 +267,27 @@ class WebSocketClient {
...
+    val req = Request.Builder().url(getWebSocketUrl()).build()
     this.websocket?.close(1000, "Reset Websocket")
     this.heartbeatFuture?.cancel(true)
-    wsListener = WebSocketListenerImpl(GPCloudOptions.websocketAuthToken, this::onAuthDone, this::onMessage, this::onClose)
+    val wsListener = WebSocketListenerImpl(GPCloudOptions.websocketAuthToken, this::onAuthDone, this::onMessage, this::onClose)
+    this.wsListener = wsListener
     this.websocket = OkHttpClient.Builder()
-      .connectionSpecs(listOf(ConnectionSpec.COMPATIBLE_TLS))
-      .build().newWebSocket(req, this.wsListener)
+      .connectionSpecs(listOf(ConnectionSpec.COMPATIBLE_TLS, ConnectionSpec.CLEARTEXT))
```
dbarashev, line 289: maybe wrap this into isLocal test as well? Because this probably means that should TLS in prod fail, we'll start sending data anyway in clear text.

ganttproject/src/main/java/biz/ganttproject/storage/cloud/GPCloudHttpImpl.kt
```diff
@@ -290,6 +304,7 @@ class WebSocketClient {
       when (it) {
         "ProjectLockStatusChange" -> fireLockStatusChange(payload)
         "ProjectChange", "ProjectRevert" -> fireContentsChange(payload)
+        "XlogReceived" -> fireXlogReceived(payload)
```
dbarashev, line 307: Just "Xlog"? It is obviously received here...

ganttproject/src/main/java/biz/ganttproject/storage/cloud/GPCloudHttpImpl.kt
```diff
@@ -350,23 +370,28 @@ class WebSocketClient {
...
+  fun onXlogReceived(listener: (ObjectNode) -> Unit): () -> Unit {
+    xlogReceivedListeners.add(listener)
+    return { xlogReceivedListeners.remove(listener) }
+  }
+
   private fun sendHeartbeat() {
-    this.websocket?.let { websocket ->
-      websocket.send("HB")
-    }
+    this.websocket?.send("HB")
```
dbarashev, line 379: Thanks for removing these useless constructions.

ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
```diff
@@ -161,6 +171,11 @@ public GanttProject(boolean isOnlyViewer) {
     ProjectStateHolderEventListener stateListener = new ProjectStateHolderEventListener(myProjectDatabase);
     addProjectEventListener(stateListener);
     getTaskManager().addTaskListener(stateListener);
+    if (isColloboqueLocalTest()) {
+      getWebSocket().register(null);
```
dbarashev, line 175: is it really needed? It seems that it only starts sending heartbeats (which colloboque is ignoring anyway)

ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
dbarashev: returning Unit from Java is ok

dbarashev: What is not so good is that this method will be triggered on literally every change in the tasks, which basically mean that every change will be sent to Colloboque as a separate log record. This may: 1) be inefficient and 2) break the [implied] transaction. Say, the user moved the project start to some new date. This will change the dates of all tasks, but from the user's perspective it is a single atomic transaction. 

This might be ok for now just for testing, but please add TODO to refactor it. We have so-called "undoable actions" which correspond to transactions from the user perspective (e.g. invoking a single action from the menu) and it is better to group xlof changes accumulated during such "transactions"

ganttproject/src/main/java/biz/ganttproject/storage/cloud/GPCloudHttpImpl.kt
npostnikova: Thanks!

ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
npostnikova: The only thing it does is connecting to the web socket for the first time. We need to do it so that it's the only place where we specify the url, right? 
Btw, heartbeats are not sent as they are initiated in `onAuthDone`

ganttproject/src/main/java/biz/ganttproject/storage/cloud/GPCloudHttpImpl.kt
npostnikova: Sorry, I don't get it
The server sends back _something_ which should contain the id of our last log it has saved. Do you hint that the name is bad? :) And why "obviously received"?..

ganttproject/src/main/java/biz/ganttproject/storage/cloud/GPCloudHttpImpl.kt
npostnikova: My pleasure :)

ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
npostnikova: Thanks for the remark! It should be definitely improved. Let's leave it as TODO for now, I suspect it requires quite a lot of investigation.

ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
```diff
@@ -880,4 +895,22 @@ public void refresh() {
     super.repaint();
   }
 
+  // TODO: Accumulate changes instead of sending it every time.
+  private Unit sendProjectStateLogs() {
+    gpLogger.debug("Sending project state logs");
+    try {
+      var logs = myProjectDatabase.fetchLogRecords(myLastReceivedXlogId.get() + 1, 10);
```
dbarashev, line 902: It seems that every new task send also the xlogs of the previously created task.

dbarashev (commented):
Please pass 0 timeout to the websocket server, otherwise it closes the socket in 5 seconds

cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/DevServer.kt
```diff
@@ -43,7 +54,8 @@ class DevServerMain : CliktCommand() {
     val updateInputChannel = Channel<InputXlog>()
     val dataSourceFactory = PostgresDataSourceFactory(pgHost, pgPort, pgSuperUser, pgSuperAuth)
     val colloboqueServer = ColloboqueServer(dataSourceFactory::createDataSource, initInputChannel, updateInputChannel)
-    ColloboqueHttpServer(port, colloboqueServer).start(NanoHTTPD.SOCKET_READ_TIMEOUT, false)
+    ColloboqueHttpServer(port, colloboqueServer).start(0, false)
```
dbarashev, line 57: it doesn't matter for the dev purposes, but setting a timeout for the regular HTTP connection makes sense, because it is not supposed to be open for long time.

ganttproject/src/main/java/net/sourceforge/ganttproject/storage/Xlog.kt
```diff
@@ -43,9 +46,29 @@ data class XlogRecord(
...
+ * @param committedTxnNum the number of committed transactions.
+ *
+ * TODO: Get rid of type. It's used for routing in `WebSocketClient::onMessage`.
+ * TODO: Provide transaction conflicts handling.
+ */
+@Serializable
+data class ServerCommitResponse(
+  val baseTxnId: String,
+  val committedTxnId: String,
+  val committedTxnNum: Int,
```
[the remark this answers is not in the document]
npostnikova: `txnNum` is the number of transactions which were consequently applied to the base. `txnId` is the resulting transaction id.
The client can send messages `{ "baseId", [txn1, txn2] }`, `{"baseId", [txn1, txn2, txn3]}`, `txnNum` helps to understand which request succeeded.

ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
```diff
@@ -899,18 +912,34 @@ public void refresh() {
...
+      gpLogger.error("Failed to send logs {}", new Object[]{}, ImmutableMap.of(), e);
     }
     return Unit.INSTANCE;
   }
 
-  private Unit fireXlogReceived(ObjectNode payload) {
-    myLastReceivedXlogId.getAndAccumulate(payload.get("id").asInt(-1), Math::max);
+  private Unit fireXlogReceived(ServerCommitResponse response) {
+    myXlogSendingLock.lock();
+    if (myLastCommittedTxnId.equals(response.getBaseTxnId())) {
```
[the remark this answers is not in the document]
npostnikova: Why? Server sends back `baseTxnId` which it's received from the client, the updated txn id  is stored in `committedTxnId`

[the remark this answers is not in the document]
dbarashev: Ah, I see. Maybe rename "myLastCommittedTxnId" to "myBaseTxnId"? This will remove some ambiguity from code.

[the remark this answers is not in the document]
dbarashev: ... and may be "committedTxnId" in the response to "new base txn id"?

ganttproject/src/main/java/net/sourceforge/ganttproject/storage/Xlog.kt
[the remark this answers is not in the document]
dbarashev: I doubt that it makes sense to send different messages with the same base ID. The log is supposed to be linear, and the client should not create two tails forked from the same base point.

If you meant that it would be a linear log, just split across a few chunks and txnNum establishes the order of those chunks then I suggest to keep the protocol more simple: if we have a few transactions sequentially applied to some base ID then we send the first one, wait until we get the response, get a new base ID, then send the second one with the new base ID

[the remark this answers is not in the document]
npostnikova: > If you meant that it would be a linear log
Yes, that's what I meant. The order is established by local xlog id. `txnNum` just tells which record to send next.

Okay, so that we never send a list of transactions, just a single one?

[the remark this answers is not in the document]
dbarashev: For now, let's send just transactions one by one and see how it goes.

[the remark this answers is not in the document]
npostnikova: Done

ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
```diff
@@ -899,18 +912,34 @@ public void refresh() {
   private Unit sendProjectStateLogs() {
     gpLogger.debug("Sending project state logs");
     try {
-      var logs = myProjectDatabase.fetchLogRecords(myLastReceivedXlogId.get() + 1, 10);
+      myXlogSendingLock.lock();
```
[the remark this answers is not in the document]
npostnikova: Done, not sure if it looks nice enough..

ganttproject/src/main/java/net/sourceforge/ganttproject/task/TaskImpl.java
```diff
@@ -736,9 +736,10 @@ public void setName(final String name) {
         @Override
         public void run() {
           myFieldChange.myEventSender = myPropertiesEventSender;
+          myFieldChange.setOldValue(TaskImpl.this.getName());
           myFieldChange.setValue(name);
           TaskImpl.this.setName(name);
-          if (taskUpdateBuilder != null) {
+          if (taskUpdateBuilder != null && myFieldChange.hasChange()) {
             taskUpdateBuilder.setName(name);
```
npostnikova, line 743: When a task is created or a user misclicked, the task name gets selected. If you clicked enter after that, a useless update transaction was created

ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
```diff
@@ -148,15 +146,36 @@ public class GanttProject extends GanttProjectBase implements ResourceView, Gant
...
+
+    TxnCommitInfo() {}
+    TxnCommitInfo(String baseTxnId, int baseTxnLocalId) {
+      this.baseTxnId = baseTxnId;
+      this.baseTxnLocalId = baseTxnLocalId;
+    }
+  }
+
+  private static class TxnCommitInfoSync extends TxnCommitInfo {
+    synchronized Boolean compareAndSetBaseTxn(String expectedTxnId, String newTxnId) {
```
dbarashev, line 163: You never check the return value of this method. Did you mean to add such code later?
Add TODO?

ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
```diff
@@ -148,15 +146,36 @@ public class GanttProject extends GanttProjectBase implements ResourceView, Gant
...
-  /**
-   * Number of transactions committed by the Colloboque server.
-   */
-  private int myCommittedTxnNum = 0;
-  /**
-   * ID of the last transaction committed by the Colloboque server.
-   */
-  private String myLastCommittedTxnId = "";
-  private final Lock myXlogSendingLock = new ReentrantLock();
+  private static class TxnCommitInfo {
```
dbarashev, line 149: In my opinion,Pair<String, Int> would suffice. Also, it looks as if this class was supposed to be immutable, but something went wrong, namely, that TxnCommitInfoSync which requires a default constructor and non-final members.

I suggest to keep the Sync class, but remove the inheritance and optionally replace TxnCommitInfo with a Pair

ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
npostnikova: Well, I think I don't want to check it
Calling it `set` wouldn't be correct. Maybe, it's better to rename it to `acceptCommittedTxn`?

ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
```diff
@@ -146,35 +148,35 @@ public class GanttProject extends GanttProjectBase implements ResourceView, Gant
...
-  private static class TxnCommitInfoSync extends TxnCommitInfo {
-    synchronized Boolean compareAndSetBaseTxn(String expectedTxnId, String newTxnId) {
-      if (baseTxnId.equals(expectedTxnId)) {
-        baseTxnId = newTxnId;
-        baseTxnLocalId++;
-        return true;
-      }
-      return false;
+    /** If `oldTxnId` is currently being hold, sets the txn ID to `newTxnId` and moves the local ID ahead by `committedNum`. */
+    synchronized void update(String oldTxnId, String newTxnId, int committedNum) {
```
dbarashev, line 164: I believe that with AtomicReference you don't need synchronized methods anymore. AtomicReference is thread-safe: https://jenkov.com/tutorials/java-util-concurrent/atomicreference.html

cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/DevServer.kt
```diff
@@ -59,5 +71,53 @@ class ColloboqueHttpServer(port: Int, private val colloboqueServer: ColloboqueSe
...
+  NanoWSD("localhost", port) {
+  override fun openWebSocket(handshake: IHTTPSession?): WebSocket {
+    return WebSocketImpl(handshake)
+  }
+
+  private class WebSocketImpl(handshake: IHTTPSession?) : WebSocket(handshake) {
+    private fun parseInputXlog(message: String): InputXlog? = try {
+      Json.decodeFromString<InputXlog>(message)
+    } catch (e: Exception) {
+      LOG.error("Failed to parse $message")
```
dbarashev, line 86: pls use placeholders for logging the input which comes from the outside (remember the recent log4j issues?) and _log the exception too_, it carries information which is important for debugging
```suggestion
      LOG.error("Failed to parse {}", message, e)
```

cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/DevServer.kt
```diff
@@ -59,5 +71,53 @@ class ColloboqueHttpServer(port: Int, private val colloboqueServer: ColloboqueSe
...
+      LOG.debug("WebSocket opened")
+    }
+
+    override fun onClose(code: WebSocketFrame.CloseCode?, reason: String?, initiatedByRemote: Boolean) {
+      LOG.debug("WebSocket closed")
+    }
+
+    override fun onMessage(message: WebSocketFrame?) {
+      try {
+        LOG.debug("Message received ${message?.textPayload}")
```
dbarashev, line 100: same about the placeholders

ganttproject/src/main/java/biz/ganttproject/storage/cloud/GPCloudHttpImpl.kt
```diff
@@ -332,6 +353,16 @@ class WebSocketClient {
     this.contentChangeListeners.forEach { it(payload) }
   }
 
+  private fun fireCommitResponseReceived(payload: ObjectNode) {
+    LOG.debug("Commit response received {}", payload)
+    try {
+      val serverResponse = Json.decodeFromString<ServerCommitResponse>(payload.toString())
+      xlogCommitResponseListeners.forEach { it(serverResponse) }
+    } catch (e: Exception) {
+      LOG.debug("Failed to parse ServerCommitResponse $payload")
```
dbarashev, line 362: placeholders and log the exception too

dbarashev (commented):
Looks good overall. A  few more comments below.

ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
npostnikova: Sure, I've missed it, sorry

cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/DevServer.kt
npostnikova: Thanks! Fixed

cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/DevServer.kt
```diff
@@ -108,14 +108,14 @@ class ColloboqueWebSocketServer(port: Int, private val colloboqueServer: Collobo
...
-        LOG.error("Failed to send response", e)
+        LOG.error("Failed to send response {}", e)
       }
     }
 
     override fun onPong(pong: WebSocketFrame?) {}
 
     override fun onException(exception: IOException?) {
-      LOG.error("WebSocket exception $exception")
+      LOG.error("WebSocket exception {}", exception = exception)
```
dbarashev, line 118:
```suggestion
      LOG.error("WebSocket exception {}", exception)
```

npostnikova: It's not happy cause the value is nullable, it wants it to be Any in this case

dbarashev: I think that it is nullable just because  the code generator could not prove the opposite when it was implementing this method. IRL it makes no sense to call onException method with null argument.

Anyway, `exception = exception` is ok too (and may be even better) but remove the placeholder then, as there is no value to put into the placeholder. Exception argument is handled separately.

ganttproject/src/main/java/biz/ganttproject/storage/cloud/GPCloudHttpImpl.kt
dbarashev: ping

dbarashev (approved):
LGTM

Issue #2096, opened 2022-04-20
Server-side Colloboque database schema 


dbarashev:
On the server-side we'll have to deal with concurrent updates of properties of the same task by different users. Such updates will be technically conflicting if all properties are stored in the same table.

The idea of resolving such conflicts is to spread the task properties over a few stored tables, and provide an updatable view for running queries and update statements as if they were in the same table.

PR #2103, opened 2022-04-25
Parse and insert project on the server side

dbarashev:
Working scenario of parsing an XML file and inserting the parsed tasks into Postgres using code written for client.

Update #2096


npostnikova (approved):
Looks good!

PR #2108, opened 2022-04-28
Tkt 2096 | Perform task insert on the server side

npostnikova:
* Perform task inserts and updates on the server side if the baseTxnId matches.
* Use lower case H2 + generate lower case jOOQ, so that queries can be executed on the server side.
* During handshakes, the client sends its projectRefid. The server initializes the project and sends back the baseTxnId.
* Use Postgres dialect in the client. With H2 dialect, table qualification is added to update requests which is incompatible with Postgres.

### DIFF
--- a/cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/ColloboqueServer.kt
+++ b/cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/ColloboqueServer.kt
@@ -36,38 +36,95 @@ import java.util.*
 import javax.sql.DataSource
 import net.sourceforge.ganttproject.storage.InitRecord
 import net.sourceforge.ganttproject.storage.InputXlog
+import net.sourceforge.ganttproject.storage.XlogRecord
+import java.time.LocalDateTime
+import java.util.concurrent.locks.ReentrantReadWriteLock
+import kotlin.concurrent.read
+import kotlin.concurrent.write
+
+private typealias ProjectRefid = String
+
+class ColloboqueServerException: Exception {
+  constructor(message: String): super(message)
+  constructor(message: String, cause: Throwable?): super(message, cause)
+}
 
 class ColloboqueServer(
   private val dataSourceFactory: (projectRefid: String) -> DataSource,
   private val initInputChannel: Channel<InitRecord>,
   private val updateInputChannel: Channel<InputXlog>) {
+  private val refidToDataSource: MutableMap<ProjectRefid, DataSource> = mutableMapOf()
+  private val refidToBaseTxnId: MutableMap<ProjectRefid, ProjectRefid> = mutableMapOf()
+  private val dataSourcesLock = ReentrantReadWriteLock()
 
-  fun init(projectRefid: String, debugCreateProject: Boolean) {
-    dataSourceFactory(projectRefid).let { ds ->
-      ds.connection.use {
-        it.createStatement().executeQuery("SELECT uid FROM Task").use { rs ->
-          while (rs.next()) {
-            println(rs.getString(1))
+  fun init(projectRefid: ProjectRefid, debugCreateProject: Boolean) {
+    try {
+      dataSourceFactory(projectRefid).let { ds ->
+        ds.connection.use {
+          it.createStatement().executeQuery("SELECT uid FROM Task").use { rs ->
+            while (rs.next()) {
+              println(rs.getString(1))
+            }
           }
-        }
-        if (debugCreateProject) {
-          DSL.using(it, SQLDialect.POSTGRES)
-            .configuration()
-            .deriveSettings { it.withRenderNameCase(RenderNameCase.LOWER) }
-            .dsl().let { dsl ->
+          if (debugCreateProject) {
+            DSL.using(it, SQLDialect.POSTGRES)
+              .configuration()
+              .deriveSettings { it.withRenderNameCase(RenderNameCase.LOWER) }
+              .dsl().let { dsl ->
 
-              loadProject("""
+                loadProject(
+                  """
 <?xml version="1.0" encoding="UTF-8"?>
 <project name="" company="" webLink="" view-date="2022-01-01" view-index="0" gantt-divider-location="374" resource-divider-location="322" version="3.0.2906" locale="en">
     <tasks empty-milestones="true">
         <task id="0" uid="qwerty" name="Task1" color="#99ccff" meeting="false" start="2022-02-10" duration="25" complete="85" expand="true"/>
     </tasks>
 </project>
-          """.trimIndent(), dsl)
-            }
+          """.trimIndent(), dsl
+                )
+              }
+          }
+          dataSourcesLock.write {
+            refidToDataSource[projectRefid] = ds
+            refidToBaseTxnId[projectRefid] = ""  // TODO: get from the database
+          }
         }
       }
+    } catch (e: Exception) {
+      throw ColloboqueServerException("Failed to init project $projectRefid", e)
     }
+  }
+
+  /** Performs transaction commit if `baseTxnId` corresponds to the value hold by the server. */
+  fun commitTxnIfSucc(projectRefid: ProjectRefid, baseTxnId: String, transaction: XlogRecord): ProjectRefid? = dataSourcesLock.read {
+    if (transaction.sqlStatements.isEmpty()) throw ColloboqueServerException("Empty transactions not allowed")
+    if (getBaseTxnId(projectRefid) != baseTxnId) return null
+    val dataSource = refidToDataSource[projectRefid]
+      ?: throw ColloboqueServerException("DataSource for project $projectRefid not set")
+    try {
+      var newBaseTxnId: String? = null
+      DSL.using(dataSource, SQLDialect.POSTGRES)
+        .transaction { config ->
+          transaction.sqlStatements.forEach { config.dsl().execute(it) }
+          newBaseTxnId = generateNextTxnId(projectRefid, baseTxnId, transaction)
+          // TODO: update transaction id in the database
+        }
+      dataSourcesLock.write {
+        refidToBaseTxnId[projectRefid] = newBaseTxnId ?: baseTxnId
+        return newBaseTxnId
+      }
+    } catch (e: Exception) {
+      throw ColloboqueServerException("Failed to commit transaction", e)
+    }
+  }
+
+  private fun getBaseTxnId(projectRefid: ProjectRefid): String = dataSourcesLock.read {
+    refidToBaseTxnId[projectRefid] ?: throw ColloboqueServerException("Project $projectRefid not initialized")
+  }
+
+  // TODO
+  private fun generateNextTxnId(projectRefid: ProjectRefid, oldTxnId: String, transaction: XlogRecord): String {
+    return LocalDateTime.now().toString()
   }
 }
 
--- a/cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/ColloboqueServer.kt
+++ b/cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/ColloboqueServer.kt
@@ -101,7 +101,7 @@ class ColloboqueServer(
   private suspend fun processUpdatesLoop() {
     for (inputXlog in updateInputChannel) {
       try {
-        val newBaseTxnId = commitTxnIfSucc(inputXlog.projectRefid, inputXlog.baseTxnId, inputXlog.transactions[0])
+        val newBaseTxnId = applyXlog(inputXlog.projectRefid, inputXlog.baseTxnId, inputXlog.transactions[0])
           ?: continue
         val response = ServerCommitResponse(
           inputXlog.baseTxnId,
@@ -127,7 +127,7 @@ class ColloboqueServer(
    * Performs transaction commit if `baseTxnId` corresponds to the value hold by the server.
    * Returns new baseTxnId on success.
    */
-  private fun commitTxnIfSucc(projectRefid: ProjectRefid, baseTxnId: String, transaction: XlogRecord): String? {
+  private fun applyXlog(projectRefid: ProjectRefid, baseTxnId: String, transaction: XlogRecord): String? {
     if (transaction.sqlStatements.isEmpty()) throw ColloboqueServerException("Empty transactions not allowed")
     if (getBaseTxnId(projectRefid) != baseTxnId) throw ColloboqueServerException("Invalid transaction id $baseTxnId")
     try {
--- a/cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/ColloboqueServer.kt
+++ b/cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/ColloboqueServer.kt
@@ -35,9 +35,9 @@ import org.jooq.DSLContext
 import org.jooq.SQLDialect
 import org.jooq.conf.RenderNameCase
 import org.jooq.impl.DSL
+import java.sql.Connection
 import java.text.DateFormat
 import java.util.*
-import javax.sql.DataSource
 import java.time.LocalDateTime
 import java.util.concurrent.Executors
 
@@ -49,7 +49,8 @@ class ColloboqueServerException: Exception {
 }
 
 class ColloboqueServer(
-  private val dataSourceFactory: (projectRefid: String) -> DataSource,
+  private val initProject: (projectRefid: String) -> Unit,
+  private val connectionFactory: (projectRefid: String) -> Connection,
   private val initInputChannel: Channel<InitRecord>,
   private val updateInputChannel: Channel<InputXlog>,
   private val serverResponseChannel: Channel<String>) {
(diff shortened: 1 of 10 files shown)


cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/ColloboqueServer.kt
```diff
@@ -36,39 +36,96 @@ import java.util.*
...
+class ColloboqueServerException: Exception {
+  constructor(message: String): super(message)
+  constructor(message: String, cause: Throwable?): super(message, cause)
+}
 
 class ColloboqueServer(
   private val dataSourceFactory: (projectRefid: String) -> DataSource,
   private val initInputChannel: Channel<InitRecord>,
   private val updateInputChannel: Channel<InputXlog>) {
+  private val refidToDataSource: MutableMap<ProjectRefid, DataSource> = mutableMapOf()
```
dbarashev, line 56: I suggest keeping this project to data source mapping in the factory, to make code in this file easier.

cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/DevServer.kt
```diff
@@ -95,20 +99,40 @@ class ColloboqueWebSocketServer(port: Int, private val colloboqueServer: Collobo
...
+          // TODO: add multiple transactions support.
+          LOG.error("Only single transaction commit supported")
+          return
+        }
+        var newBaseTxnId: String? = null
+        try {
+          newBaseTxnId =
+            colloboqueServer.commitTxnIfSucc(inputXlog.projectRefid, inputXlog.baseTxnId, inputXlog.transactions[0])
+        } catch (e: Exception) {
+          LOG.error("Failed to commit\n {}", inputXlog, e)
```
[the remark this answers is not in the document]
dbarashev: It seems that it can be configured in the logback encoder pattern:
https://nurkiewicz.com/2011/09/logging-exceptions-root-cause-first.html

ganttproject/src/main/java/net/sourceforge/ganttproject/storage/Xlog.kt
```diff
@@ -69,4 +69,16 @@ data class ServerCommitResponse(
...
\ No newline at end of file
+)
+
+/** Response from the server that signals about transaction commit failure. */
+@Serializable
+data class ServerCommitError(
+  val baseTxnId: String,
+  val projectRefid: String,
+  val message: String,
+  val type: String
```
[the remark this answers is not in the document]
npostnikova: Do we want to parse it [here](https://github.com/bardsoftware/ganttproject/blob/2356f974ae26682a7f89b495c643226f438622af/ganttproject/src/main/java/biz/ganttproject/storage/cloud/GPCloudHttpImpl.kt#L321)?
I suspect that we could match with `ColloboqueResponseType.COMMIT_SUCCESS.name` or so, but it doesn't look robust

cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/ColloboqueServer.kt
```diff
@@ -36,39 +36,96 @@ import java.util.*
...
+              }
+          }
+          dataSourcesLock.write {
+            refidToDataSource[projectRefid] = ds
+            refidToBaseTxnId[projectRefid] = "abacaba"  // TODO: get from the database
+          }
         }
       }
+    } catch (e: Exception) {
+      throw ColloboqueServerException("Failed to init project $projectRefid", e)
```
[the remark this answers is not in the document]
npostnikova: Don't you mix it with logging? :)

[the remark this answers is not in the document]
dbarashev: Ah damn :)

ganttproject/src/main/java/net/sourceforge/ganttproject/storage/Xlog.kt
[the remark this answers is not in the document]
dbarashev: I meant that there might be a built-in support of enums in Kotlin serialization. If enums are not supported, let's leave as is.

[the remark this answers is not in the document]
npostnikova: They are supported but in `onMessage` we take the text value. Do you want GPServer to send enums too? (`ProjectChange` etc)

cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/DevServer.kt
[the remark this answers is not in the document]
npostnikova: Nice feature! But the main problem is that the stacktrace is not printed at all, and I don't see why

cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/DevServer.kt
```diff
@@ -73,13 +73,36 @@ class ColloboqueHttpServer(port: Int, private val colloboqueServer: ColloboqueSe
...
+  private class WebSocketImpl(handshake: IHTTPSession,
+                              private val colloboqueServer: ColloboqueServer,
+                              private val updateInputChannel: Channel<InputXlog>,
+                              private val serverResponseChannel: Channel<String>) : WebSocket(handshake) {
+    init {
+      handshake.parameters["projectRefid"]?.firstOrNull()?.also { refid ->
+        colloboqueServer.init(refid, false)
+        this.handshakeResponse.addHeader("baseTxnId", colloboqueServer.getBaseTxnId(refid))
+      }
+      GlobalScope.launch {
```
dbarashev, line 94: I suggest creating an individual scope for sending the responses (and actually for other tasks as well, as it provides more control), e.g. like this:

```
private val responseScope = CoroutineScope(Executors.newSingleThreadExecutor().asCoroutineDispatcher())
...
responseScope.launch {}
```

cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/Postgres.kt
```diff
@@ -33,6 +33,7 @@ class PostgresDataSourceFactory(
     jdbcUrl = "jdbc:postgresql://${pgHost}:${pgPort}/dev_all_projects"
   }
   private val superDataSource = HikariDataSource(superConfig)
+  private val cachedDataSources: MutableMap<ProjectRefid, DataSource> = mutableMapOf()
```
dbarashev, line 36: I don't think it makes too much sense now because we're using the same database for all projects anyway. It will make sense later, if and when we start using different databases. But as soon as there is just one database, one data source should be enough

ganttproject/src/main/java/biz/ganttproject/storage/cloud/GPCloudHttpImpl.kt
```diff
@@ -363,6 +372,15 @@ class WebSocketClient {
     }
   }
 
+  // TODO: Propagate info to the client so that they could resolve conflicts.
+  private fun fireCommitErrorReceived(payload: ObjectNode) {
+    LOG.debug("Commit error received:\n {}", payload)
```
dbarashev, line 377: s/debug/error` ?

cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/ColloboqueServer.kt
```diff
@@ -23,52 +23,134 @@ import biz.ganttproject.core.io.parseXmlProject
...
 import javax.sql.DataSource
-import net.sourceforge.ganttproject.storage.InitRecord
-import net.sourceforge.ganttproject.storage.InputXlog
+import java.time.LocalDateTime
 
+internal typealias ProjectRefid = String
+
+class ColloboqueServerException: Exception {
+  constructor(message: String): super(message)
+  constructor(message: String, cause: Throwable?): super(message, cause)
```
dbarashev, line 49: why one may want to pass a `null` cause?

cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/ColloboqueServer.kt
```diff
@@ -23,52 +23,134 @@ import biz.ganttproject.core.io.parseXmlProject
...
+        )
+        serverResponseChannel.send(Json.encodeToString(errorResponse))
+      }
+    }
+  }
+
+  /** Performs transaction commit if `baseTxnId` corresponds to the value hold by the server. */
+  private fun commitTxnIfSucc(projectRefid: ProjectRefid, baseTxnId: String, transaction: XlogRecord): ProjectRefid? {
+    if (transaction.sqlStatements.isEmpty()) throw ColloboqueServerException("Empty transactions not allowed")
+    if (getBaseTxnId(projectRefid) != baseTxnId) return null
```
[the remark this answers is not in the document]
npostnikova: Isn't it a normal case which can be ignored? A client could send multiple requests with the same update (either because few events happened or we do it periodically), the first one is processed, others can be just skipped

cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/ColloboqueServer.kt
```diff
@@ -23,52 +23,134 @@ import biz.ganttproject.core.io.parseXmlProject
...
+          refidToBaseTxnId[projectRefid] = "abacaba"  // TODO: get from the database
         }
       }
+    } catch (e: Exception) {
+      throw ColloboqueServerException("Failed to init project $projectRefid", e)
     }
   }
+
+  private suspend fun processUpdatesLoop() {
+    while (true) {
```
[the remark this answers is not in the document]
npostnikova: Thanks!!

cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/ColloboqueServer.kt
```diff
@@ -23,52 +23,134 @@ import biz.ganttproject.core.io.parseXmlProject
...
+          e.message.orEmpty(),
+          SERVER_COMMIT_ERROR_TYPE
+        )
+        serverResponseChannel.send(Json.encodeToString(errorResponse))
+      }
+    }
+  }
+
+  /** Performs transaction commit if `baseTxnId` corresponds to the value hold by the server. */
+  private fun commitTxnIfSucc(projectRefid: ProjectRefid, baseTxnId: String, transaction: XlogRecord): ProjectRefid? {
```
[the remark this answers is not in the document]
npostnikova: Right, it should be just a `String`, ty

cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/Postgres.kt
npostnikova: Do you mean changing the schema every time the database is accessed?

ganttproject/src/main/java/biz/ganttproject/storage/cloud/GPCloudHttpImpl.kt
npostnikova: Changed to error

cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/ColloboqueServer.kt
npostnikova: Thanks, done :man_facepalming:

cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/DevServer.kt
npostnikova: Do you mean having an individual scope per instance (rather then GlobalScope) or per each need (sending/receiving)?

cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/ColloboqueServer.kt
```diff
@@ -23,52 +23,134 @@ import biz.ganttproject.core.io.parseXmlProject
...
+    if (transaction.sqlStatements.isEmpty()) throw ColloboqueServerException("Empty transactions not allowed")
+    if (getBaseTxnId(projectRefid) != baseTxnId) return null
+    try {
+      var newBaseTxnId: String? = null
```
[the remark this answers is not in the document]
npostnikova: Much better now, thank you!

cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/ColloboqueServer.kt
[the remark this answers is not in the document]
dbarashev: Nope. What are possible client side strategies with respect to ? 
1. "Synchronized" request-response: a) generate UPDATE... > b) send Xlog > c) pause sending until commit response comes > d) update base txn id > e) send new updates, if any, with the new base txn id. If everything is ok, it sends a new update with a correct base txn id, but should it happen to be different, the client needs to know, because it has to resolve the issue and send a new update again
2. "Asynchronous" request-response: a) generate UPDATE... > b) send Xlog > c) generate UPDATE ... > d) send Xlog again > e) receive commit response to "b)". Here at the moment d) we still can't just ignore the Xlog, because the client needs to send it again

cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/DevServer.kt
dbarashev: individual scope for each activity, one for receiving, one for sending.

cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/Postgres.kt
dbarashev: We will do a lot of schema tricks in the normal mode anyway, even if we have one database per project. When there are many projects in the same database, we can keep their actual versions in different schemas too.

cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/Postgres.kt
```diff
@@ -57,11 +54,11 @@ class PostgresDataSourceFactory(
...
-      jdbcUrl = "jdbc:postgresql://${pgHost}:${pgPort}/dev_all_projects"
-      this.schema = schema
-    }).also { cachedDataSources[projectRefid] = it }
   }
+
+  fun createConnection(projectRefid: String): Connection =
+    superDataSource.connection.also { it.schema = getSchema(projectRefid) }
+
+  // TODO: escape projectRefid
+  private fun getSchema(projectRefid: String) = "project_$projectRefid"
```
dbarashev, line 63: refid in our reality is a pretty long string. I suggest hashing it like this:

`Hashing.murmur3_128().hashBytes(refid).toString()`

cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/Postgres.kt
```diff
@@ -57,11 +54,11 @@ class PostgresDataSourceFactory(
...
     }
-    return HikariDataSource(HikariConfig().apply {
-      username = pgSuperUser
-      password = pgSuperAuth
-      jdbcUrl = "jdbc:postgresql://${pgHost}:${pgPort}/dev_all_projects"
-      this.schema = schema
-    }).also { cachedDataSources[projectRefid] = it }
   }
+
+  fun createConnection(projectRefid: String): Connection =
```
dbarashev, line 59: super data source was intended for metadata manipulations by a super-user, such as storing instance-level mappings of projects to databases, or cloning a template database/template schema. I suggest using a separate "non-super" datasource for the regular connections

cloud.ganttproject.colloboque/src/main/kotlin/cloud/ganttproject/colloboque/ColloboqueServer.kt
```diff
@@ -129,10 +129,10 @@ class ColloboqueServer(
    */
   private fun commitTxnIfSucc(projectRefid: ProjectRefid, baseTxnId: String, transaction: XlogRecord): String? {
     if (transaction.sqlStatements.isEmpty()) throw ColloboqueServerException("Empty transactions not allowed")
-    if (getBaseTxnId(projectRefid) != baseTxnId) return null
+    if (getBaseTxnId(projectRefid) != baseTxnId) throw ColloboqueServerException("Invalid transaction id $baseTxnId")
     try {
       return DSL
-        .using(dataSourceFactory(projectRefid), SQLDialect.POSTGRES)
+        .using(connectionFactory(projectRefid), SQLDialect.POSTGRES)
```
dbarashev, line 135: Who will close the connection? I am not sure that DSLContext will do it.

dbarashev (approved):
LGTM. Good job!

PR #2110, opened 2022-04-29
XLOG prefix and base64

dbarashev:
* Added XLOG prefix to xlog messages and base64-encoded the payload
* switch on Colloboque dev mode on the client using -Pdevws flag ro gradle run command


npostnikova (approved):
:+1:

PR #2112, opened 2022-04-30
Add multi-statement transaction support

npostnikova:
* Add `startTransaction` & `commitTransaction` methods to `ProjectDatabase`
* Extend `LogRecord` with `txnId`

Note: the changes don't affect `TaskManager` yet

### DIFF
--- a/ganttproject/src/main/java/net/sourceforge/ganttproject/storage/SqlProjectDatabaseImpl.kt
+++ b/ganttproject/src/main/java/net/sourceforge/ganttproject/storage/SqlProjectDatabaseImpl.kt
@@ -25,6 +25,7 @@ import biz.ganttproject.core.time.TimeDuration
 import biz.ganttproject.storage.db.Sequences.TXNID
 import biz.ganttproject.storage.db.Tables.*
 import biz.ganttproject.storage.db.tables.records.TaskRecord
+import net.sourceforge.ganttproject.GPLogger
 import net.sourceforge.ganttproject.storage.ProjectDatabase.*
 import net.sourceforge.ganttproject.task.Task
 import net.sourceforge.ganttproject.task.TaskInfo
@@ -52,7 +53,7 @@ class SqlProjectDatabaseImpl(private val dataSource: DataSource) : ProjectDataba
 
   private class Query(
     val errorMessage: () -> String,
-    val queryBuilder: (DSLContext) -> String
+    val sqlStatement: String
   )
 
   /** Queries which belong to the current transaction. Null if each statement should be committed separately. */
@@ -83,14 +84,14 @@ class SqlProjectDatabaseImpl(private val dataSource: DataSource) : ProjectDataba
       val nextTxnId = context.nextval(TXNID)
       queries.forEach {
         try {
-          val query = it.queryBuilder(context)
-          context.execute(query)
+          context.execute(it.sqlStatement)
           context
             .insertInto(LOGRECORD)
             .set(LOGRECORD.TXN_ID, nextTxnId.toInt())
-            .set(LOGRECORD.SQL_STATEMENT, query)
+            .set(LOGRECORD.SQL_STATEMENT, it.sqlStatement)
             .execute()
         } catch (e: Exception) {
+          LOG.error("Failed to execute or log txnId={}\n {}", nextTxnId, it.sqlStatement)
           throw ProjectDatabaseException(it.errorMessage(), e)
         }
       }
@@ -99,7 +100,7 @@ class SqlProjectDatabaseImpl(private val dataSource: DataSource) : ProjectDataba
 
   /** Add a query to the current txn. Executes immediately if no transaction started. */
   private fun withLog(errorMessage: () -> String, buildQuery: (dsl: DSLContext) -> String) {
-    val query = Query(errorMessage, buildQuery)
+    val query = Query(errorMessage, buildQuery(DSL.using(SQLDialect.POSTGRES)))
     currentTxn?.add(query) ?: executeAndLog(listOf(query))
   }
 
@@ -265,3 +266,5 @@ private fun Task.logId(): String = "${uid}:${taskID}"
 
 private const val H2_IN_MEMORY_URL = "jdbc:h2:mem:gantt-project-state;DB_CLOSE_DELAY=-1"
 private const val DB_INIT_SCRIPT_PATH = "/sql/init-project-database.sql"
+
+private val LOG = GPLogger.create("ProjectDatabase")
--- a/ganttproject/src/main/java/net/sourceforge/ganttproject/storage/SqlProjectDatabaseImpl.kt
+++ b/ganttproject/src/main/java/net/sourceforge/ganttproject/storage/SqlProjectDatabaseImpl.kt
@@ -175,9 +175,7 @@ class SqlTaskUpdateBuilder(private val task: Task,
     TODO("Not yet implemented")
   }
 
-  override fun setStart(start: GanttCalendar?) {
-    TODO("Not yet implemented")
-  }
+  override fun setStart(start: GanttCalendar) = nextStep { it.set(TASK.START_DATE, start.toLocalDate()) }
 
   override fun setEnd(end: GanttCalendar?) {
     TODO("Not yet implemented")
--- a/ganttproject/src/main/java/net/sourceforge/ganttproject/storage/SqlProjectDatabaseImpl.kt
+++ b/ganttproject/src/main/java/net/sourceforge/ganttproject/storage/SqlProjectDatabaseImpl.kt
@@ -38,7 +38,6 @@ import org.jooq.conf.ParamType
 import org.jooq.impl.DSL
 import java.awt.Color
 import java.sql.SQLException
-import java.util.stream.Collectors
 import javax.sql.DataSource
 import kotlin.text.Charsets
 
@@ -50,6 +49,14 @@ class SqlProjectDatabaseImpl(private val dataSource: DataSource) : ProjectDataba
       return SqlProjectDatabaseImpl(dataSource)
     }
   }
+
+  private class Query(
+    val errorMessage: () -> String,
+    val queryBuilder: (DSLContext) -> String
+  )
+
+  /** Queries which belong to the current transaction. Null if each statement should be committed separately. */
+  private var currentTxn: MutableList<Query>? = null
 
   private fun <T> withDSL(
     errorMessage: () -> String = { "Failed to execute query" },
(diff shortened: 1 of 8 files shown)


ganttproject/src/main/java/net/sourceforge/ganttproject/storage/SqlProjectDatabaseImpl.kt
```diff
@@ -127,34 +148,51 @@ class SqlProjectDatabaseImpl(private val dataSource: DataSource) : ProjectDataba
...
-      .where(LOGRECORD.ID.ge(startId))
-      .orderBy(LOGRECORD.ID.asc())
-      .limit(limit)
-      .map { LogRecord(it.id, it.sqlStatement) }
+      .where(LOGRECORD.TXN_ID.ge(startTxnId).and(LOGRECORD.TXN_ID.lt(startTxnId + limit)))
+      .orderBy(LOGRECORD.TXN_ID, LOGRECORD.ID)
+      .fetchGroups(LOGRECORD.TXN_ID, LOGRECORD.SQL_STATEMENT)
+      .toList()
+      .sortedBy { it.first }
+      .map { XlogRecord(it.second) }
```
dbarashev, line 174: I don't think that you need to apply this toList and subsequent sorting. It is not going to be expensive, but makes the idea more difficult to understand. If you want to map the collection of list of statements, which is produced by fetchGroups to a list of XlogRecord, you can use `.values.map { XlogRecord(it) }`

ganttproject/src/main/java/net/sourceforge/ganttproject/storage/SqlProjectDatabaseImpl.kt
```diff
@@ -67,22 +77,33 @@ class SqlProjectDatabaseImpl(private val dataSource: DataSource) : ProjectDataba
...
-        .insertInto(LOGRECORD)
-        .set(LOGRECORD.SQL_STATEMENT, query)
-        .execute()
+      val nextTxnId = context.nextval(TXNID)
+      queries.forEach {
+        try {
+          context.execute(it.sqlStatement)
+          context
+            .insertInto(LOGRECORD)
+            .set(LOGRECORD.TXN_ID, nextTxnId.toInt())
```
dbarashev, line 90: It appears that you're mixing the locally generated sequence of numbers with the txn id from the server response. This is not a good idea, because they are different. In the presence of concurrent updates the server chooses the order of committing and the server generates the transaction ids. 

Local ids are also important, they will allow us to establish a reliable order of the xlog records, but they shall not be mixed with the server base txn ids.

ganttproject/src/main/resources/resources/sql/init-project-database.sql
```diff
@@ -34,7 +34,10 @@ create table if not exists TaskDependency (
 
 create table if not exists LogRecord (
   id                identity    not null,
+  txn_id            integer     not null,
```
dbarashev, line 37: Rename this to `local_txn_id` to avoid mess with the base tn id

ganttproject/src/main/java/net/sourceforge/ganttproject/storage/SqlProjectDatabaseImpl.kt
npostnikova: That's nicer, thanks!

ganttproject/src/main/java/net/sourceforge/ganttproject/storage/SqlProjectDatabaseImpl.kt
npostnikova: It seems to be fine as the local ids are never accessed for read, doesn't it? They only specify the order of transactions

ganttproject/src/main/java/net/sourceforge/ganttproject/storage/SqlProjectDatabaseImpl.kt
```diff
@@ -127,37 +154,58 @@ class SqlProjectDatabaseImpl(private val dataSource: DataSource) : ProjectDataba
...
-      lastSetStep?.let { updateQuery ->
-        executeQuery(updateQuery.where(TASK.UID.eq(task.uid)).getSQL(ParamType.INLINED))
+      lastSetStepH2?.let { updateQueryH2 ->
+        onCommit(updateQueryH2.where(TASK.UID.eq(task.uid)).getSQL(ParamType.INLINED),
+          lastSetStepPostgres?.where(TASK.UID.eq(task.uid))?.getSQL(ParamType.INLINED)
+            ?: error("No update statement for postgres"))
       }
     } finally {
-      lastSetStep = null
+      lastSetStepH2 = null
```
dbarashev, line 207: is this builder supposed to be used after commit? If not, you can remove this nullification

ganttproject/src/main/resources/resources/sql/init-project-database.sql
```diff
@@ -34,7 +34,10 @@ create table if not exists TaskDependency (
 
 create table if not exists LogRecord (
   id                identity    not null,
+  local_txn_id      integer     not null,
   sql_statement     varchar     not null,
 
   primary key (id)
 );
+
+create sequence TxnId start with 1;
```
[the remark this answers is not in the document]
dbarashev: Well, it is not really necessary to use sequences. Perhaps non-DB counter (just a simple AtomicInteger) is enough, if it is okay that the sequence starts with 1 every time when GanttProject starts and terminates when it closes.

ganttproject/src/main/java/net/sourceforge/ganttproject/storage/SqlProjectDatabaseImpl.kt
```diff
@@ -127,37 +154,58 @@ class SqlProjectDatabaseImpl(private val dataSource: DataSource) : ProjectDataba
...
+    lastSetStepPostgres = step(lastSetStepPostgres ?: DSL.using(SQLDialect.POSTGRES).update(TASK))
   }
 
   @Throws(ProjectDatabaseException::class)
-  override fun execute() {
+  override fun commit() {
     try {
-      lastSetStep?.let { updateQuery ->
-        executeQuery(updateQuery.where(TASK.UID.eq(task.uid)).getSQL(ParamType.INLINED))
+      lastSetStepH2?.let { updateQueryH2 ->
```
dbarashev, line 201: These constructions look difficult to read. What about rewriting them this way?

```
      val finalH2 = lastSetStepH2?.where(TASK.UID.eq(task.uid))?.getSQL(ParamType.INLINED) ?: error("...")  
      val finalPostgres = ....
      onCommit(finalH2, finalPostgres)
```

dbarashev (commented):
Mostly looks good. Some comments on code readability.

ganttproject/src/main/resources/resources/sql/init-project-database.sql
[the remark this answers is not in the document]
npostnikova: The charm of using sequences is that they shouldn't be incremented if a transaction rolled back
When we use `fetchTransactions` to send a txn to the websocket, we rely on the sequential order of `local_txn_id` at the moment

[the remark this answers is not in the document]
dbarashev: This is not true on Postgres and I am almost sure that on H2 sequence values are never re-used and do not rollback. It is the sequence contract that they do not generate the same number twice.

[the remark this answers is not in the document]
npostnikova: Oh, okay, that sounds reasonable
Added a local counter that is incremented only if a commit didn't fail

ganttproject/src/main/java/net/sourceforge/ganttproject/storage/SqlProjectDatabaseImpl.kt
```diff
@@ -67,22 +78,38 @@ class SqlProjectDatabaseImpl(private val dataSource: DataSource) : ProjectDataba
...
+  /** Add a query to the current txn. Executes immediately if no transaction started. */
+  private fun withLog(errorMessage: () -> String, buildQuery: (dsl: DSLContext) -> String) {
+    val query = Query(errorMessage, buildQuery(DSL.using(SQLDialect.H2)), buildQuery(DSL.using(SQLDialect.POSTGRES)))
+    currentTxn?.add(query) ?: executeAndLog(listOf(query))
+  }
+
+  private fun withLog(h2Query: String, postgresQuery: String, errorMessage: () -> String) {
+    val query = Query(errorMessage, h2Query, postgresQuery)
+    currentTxn?.add(query) ?: executeAndLog(listOf(query))
+  }
```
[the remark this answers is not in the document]
npostnikova: My intention was to place the lambda as the last argument :) Moved it back
And I don't know if we really need these logs... Everything can be found in the stacktrace, but queries are quite long. Maybe it's not bad to see a short reason with task id etc

dbarashev:
It is now complaining as follows:

```
 InputXlog(baseTxnId=abacaba, userId=userId, projectRefid=projectRefid, transactions=[XlogRecord(sqlStatements=[insert into "task" ("uid", "num", "name", "color", "shape", "is_milestone", "is_project_task", "start_date", "duration", "completion", "earliest_start_date", "priority", "web_link", "cost_manual_value", "is_cost_calculated", "notes") values ('ce8d1ac048464025aaabbf449500480a', 0, 'task_0', null, null, false, false, date '2022-05-04', 1, 0, null, '1', null, null, null, null)])])
cloud.ganttproject.colloboque.ColloboqueServerException: Project projectRefid not initialized
        at cloud.ganttproject.colloboque.ColloboqueServer.getBaseTxnId(ColloboqueServer.kt:150)
        at cloud.ganttproject.colloboque.ColloboqueServer.applyXlog(ColloboqueServer.kt:132)
        at cloud.ganttproject.colloboque.ColloboqueServer.processUpdatesLoop(ColloboqueServer.kt:104)
        at cloud.ganttproject.colloboque.ColloboqueServer.access$processUpdatesLoop(ColloboqueServer.kt:51)
        at cloud.ganttproject.colloboque.ColloboqueServer$processUpdatesLoop$1.invokeSuspend(ColloboqueServer.kt)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
        at java.base/java.lang.Thread.run(Thread.java:833)
```

Not sure if it needs to be fixed in this PR, though.

dbarashev (approved):
LGTM, but see the last comment: transactions fail because project is not initialized.

npostnikova:
> Project projectRefid not initialized

Thanks! During merge, I accidentally overwrote the sample projectRefid used in GanttProject. It had been replaced with `refid` earlier so that it could be used for schema name without any modification.

dbarashev (approved):
LGTM. Merge it!

Issue #2083, opened 2022-04-11
Store the project model in a relational database


dbarashev:
This is a summary issue for all tasks related to the storage of project model in a relational database

### Overview
We plan to piggy-back on the relational database features, such as transactions, to implement real-time sync between the Desktop and Cloud, and merge offline Desktop data with the changes made in the Cloud. The first thing which needs to be done for this is storing and updating the project model in an in-memory project database.

PR #2118, opened 2022-05-04
API for running queries against task table

dbarashev:
Update #2083


ganttproject/src/main/java/net/sourceforge/ganttproject/storage/ProjectDatabase.kt
```diff
@@ -70,4 +70,10 @@ interface ProjectDatabase {
   /** Fetch transactions starting with the specified transaction id. */
   @Throws(ProjectDatabaseException::class)
   fun fetchTransactions(startTxnId: Int = 0, limit: Int): List<XlogRecord>
+
+  /** Run a query with the given `whenExpression` against the Task table.
```
npostnikova, line 74: > `whenExpression`

Typo

npostnikova (approved):
Looks good!

ganttproject/src/main/java/net/sourceforge/ganttproject/storage/ProjectDatabase.kt
dbarashev: thanks!

Issue #150, opened 2015-06-28
Allow simple formulas in the custom columns


GoogleCodeExporter:
```
Copied from:

https://sourceforge.net/tracker/index.php?func=detail&aid=3075106&group_id=72728
&atid=535430


```

Original issue reported on code.google.com by `dbarashev` on 3 Oct 2010 at 6:22

PR #2119, opened 2022-05-06
UI and persistence for calculated property definitions

dbarashev:
One can create calculated custom properties and persist them to the project file. No real calculations yet.


dbarashev:
Update  #150

PR #2141, opened 2022-06-05
Fix customPropertiesChange::hasChange

npostnikova:
Override `equals` for `CustomColumnsValues` to use `HashMap::equals`. It prevents from generating SQL statements for custom properties when no changes were made.

### DIFF
--- a/ganttproject/src/main/java/biz/ganttproject/customproperty/CustomColumnsValues.java
+++ b/ganttproject/src/main/java/biz/ganttproject/customproperty/CustomColumnsValues.java
@@ -162,12 +162,12 @@ public class CustomColumnsValues implements CustomPropertyHolder, Cloneable {
   public boolean equals(Object object) {
     if (this == object) return true;
     if (!(object instanceof CustomColumnsValues)) return false;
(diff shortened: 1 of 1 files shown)


ganttproject/src/main/java/biz/ganttproject/customproperty/CustomColumnsValues.java
```diff
@@ -160,4 +157,17 @@ static String getValueAsString(Object value) {
     }
     return result;
   }
+
+  @Override
+  public boolean equals(Object object) {
+    if (this == object) return true;
+    if (!(object instanceof CustomColumnsValues)) return false;
+    CustomColumnsValues columnsValues = (CustomColumnsValues) object;
+    return mapCustomColumnValue.equals(columnsValues.mapCustomColumnValue) && myManager.equals(columnsValues.myManager);
```
npostnikova, line 166: Seems like the manager is never set to null, is that correct? Or should I add and extra check for that?

dbarashev: I don't think that it is needed at all. Just remove it from equals.

dbarashev (approved):

PR #2159, opened 2022-06-13
Add LogRecord table update callbacks

npostnikova:
* Trigger txn sending to Colloboque when a new log record is inserted
* Send txns sequentially

### DIFF
--- a/ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
+++ b/ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
@@ -86,6 +86,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 
@@ -208,10 +209,6 @@ public class GanttProject extends GanttProjectBase implements ResourceView, Gant
       getWebSocket().register(null);
       getWebSocket().onCommitResponseReceived(this::fireXlogReceived);
       getWebSocket().onBaseTxnIdReceived(this::onBaseTxnIdReceived);
-      var taskListenerAdapter = new TaskListenerAdapter();
-      // TODO: add listeners sensibly.
-      taskListenerAdapter.setTaskAddedHandler(event -> this.sendProjectStateLogs());
-      getTaskManager().addTaskListener(taskListenerAdapter);
     }
 
     area = new GanttGraphicArea(this, getTaskManager(), getZoomManager(), getUndoManager(),
@@ -956,7 +953,8 @@ public class GanttProject extends GanttProjectBase implements ResourceView, Gant
     super.repaint();
   }
 
-  // TODO: Accumulate changes instead of sending it every time.
+  private final AtomicBoolean isSendingInProgress = new AtomicBoolean();
+
   private Unit sendProjectStateLogs() {
     gpLogger.debug("Sending project state logs");
     try {
@@ -969,20 +967,34 @@ public class GanttProject extends GanttProjectBase implements ResourceView, Gant
           "refid",
           txns
         ));
+        isSendingInProgress.set(true);
       }
     } catch (ProjectDatabaseException e) {
       gpLogger.error("Failed to send logs", new Object[]{}, ImmutableMap.of(), e);
+      isSendingInProgress.set(false);
+    }
+    return Unit.INSTANCE;
+  }
+
+  @Override
+  protected Unit onProjectLogUpdate() {
+    if (isColloboqueLocalTest()) {
+      super.onProjectLogUpdate();
+      if (!isSendingInProgress.get()) sendProjectStateLogs();
     }
     return Unit.INSTANCE;
   }
 
   private Unit fireXlogReceived(ServerCommitResponse response) {
     myBaseTxnCommitInfo.update(response.getBaseTxnId(), response.getNewBaseTxnId(), 1);
+    isSendingInProgress.set(false);
+    sendProjectStateLogs();
     return Unit.INSTANCE;
   }
 
   private Unit onBaseTxnIdReceived(String baseTxnId) {
     myBaseTxnCommitInfo.update("", baseTxnId, 0);
+    sendProjectStateLogs();
     return Unit.INSTANCE;
   }
 }
--- a/ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProjectBase.java
+++ b/ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProjectBase.java
@@ -40,6 +40,7 @@ import javafx.beans.property.SimpleDoubleProperty;
 import javafx.beans.property.SimpleIntegerProperty;
 import javafx.beans.property.SimpleObjectProperty;
 import javafx.collections.FXCollections;
+import kotlin.Unit;
 import kotlin.jvm.functions.Function0;
 import net.sourceforge.ganttproject.chart.Chart;
 import net.sourceforge.ganttproject.chart.ChartModelBase;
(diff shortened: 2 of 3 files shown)


ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
```diff
@@ -969,20 +967,34 @@ private Unit sendProjectStateLogs() {
           "refid",
           txns
         ));
+        isSendingInProgress.set(true);
```
dbarashev, line 970: set the guarding flag before you actually send, otherwise there is a chance to set it to true when the response is already received

ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
```diff
@@ -969,20 +967,34 @@ private Unit sendProjectStateLogs() {
...
+    if (isColloboqueLocalTest()) {
+      super.onProjectLogUpdate();
+      if (!isSendingInProgress.get()) sendProjectStateLogs();
     }
     return Unit.INSTANCE;
   }
 
   private Unit fireXlogReceived(ServerCommitResponse response) {
     myBaseTxnCommitInfo.update(response.getBaseTxnId(), response.getNewBaseTxnId(), 1);
+    isSendingInProgress.set(false);
```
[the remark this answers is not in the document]
npostnikova: Thank you so much for the idea! I hated this AtomicBoolean but I couldn't come up with something better
Not sure if I implemented this exactly as expected though...

ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
```diff
@@ -969,20 +967,34 @@ private Unit sendProjectStateLogs() {
...
+      isSendingInProgress.set(false);
+    }
+    return Unit.INSTANCE;
+  }
+
+  @Override
+  protected Unit onProjectLogUpdate() {
+    if (isColloboqueLocalTest()) {
+      super.onProjectLogUpdate();
+      if (!isSendingInProgress.get()) sendProjectStateLogs();
```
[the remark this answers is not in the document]
npostnikova: It was designed this weird because `fireXlogReceived` & `onBaseTxnIdReceived` reset the flag

ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
```diff
@@ -956,33 +951,62 @@ public void refresh() {
...
+    sendProjectStateLogs();
     return Unit.INSTANCE;
   }
 
   private Unit onBaseTxnIdReceived(String baseTxnId) {
     myBaseTxnCommitInfo.update("", baseTxnId, 0);
+    var listener = txnSendingListener.get();
+    // Websocket is [re-]started. Previous messages are discarded.
+    if (listener != null) listener.onSendCompleted();
+    sendProjectStateLogs();
```
dbarashev, line 1009: It appears that the comment above (previous messages are discarded) and these two lines (sendCompleted, sendLogs) are kinda controversial. We obviously need to make sure that the state which we have locally matches the server state which corresponds to this `baseTxnId`. Until it is not done, we probably want to really discard the changes, that is, clear the state, whatever it is, but not do what we do in the normal success case.

ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
```diff
@@ -956,33 +951,62 @@ public void refresh() {
...
       }
     } catch (ProjectDatabaseException e) {
       gpLogger.error("Failed to send logs", new Object[]{}, ImmutableMap.of(), e);
     }
     return Unit.INSTANCE;
   }
 
+  @Override
+  protected Unit onProjectLogUpdate() {
+    if (isColloboqueLocalTest()) {
```
dbarashev, line 990: I don't like that we have more and more these checks. I thought that we would have just a few differences between the "local test" and "prod" modes (url, authentication), but now it seems that we will just not do anything in prod.

ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
```diff
@@ -956,33 +951,62 @@ public void refresh() {
...
+          getWebSocket().sendLogs(new InputXlog(
+            baseTxnCommitInfo.left,
+            "userId",
+            "refid",
+            txns
+          ));
+        }
       }
     } catch (ProjectDatabaseException e) {
       gpLogger.error("Failed to send logs", new Object[]{}, ImmutableMap.of(), e);
```
dbarashev, line 983: If we are sure that `ProjectDatabaseException` will not be thrown when listener is set then maybe move the body of `if` out of try?

ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
```diff
@@ -956,33 +951,62 @@ public void refresh() {
...
-          "refid",
-          txns
-        ));
+        var listener = new TxnSendListener() {
+          @Override
+          public void onSendCompleted() {
+            txnSendingListener.compareAndSet(this, null);
+          }
+        };
+        if (txnSendingListener.compareAndSet(null, listener)) {
```
dbarashev, line 973: `else` block is missing. If it is just okay to ignore then write an empty `else` block with the explicit comment saying that it is ok. Otherwise, if failing `compareAndSet` here is not something that we expect, I suggest logging a warning (at least)

ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
npostnikova: Sounds good, done

ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
npostnikova: Thanks!
Added a TODO for performing synchronization with the server. For now, logs are sent only when the server specifies that the project is empty.

ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
dbarashev: remove this check? And maybe the super call too?

ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
```diff
@@ -956,33 +955,73 @@ public void refresh() {
...
 
+  // TODO: sync logs with the server.
   private Unit onBaseTxnIdReceived(String baseTxnId) {
-    myBaseTxnCommitInfo.update("", baseTxnId, 0);
+    // Websocket is [re-]started. Previous messages are discarded.
+    txnSendingListener.set(null);
+    if (EMPTY_LOG_BASE_TXN_ID.equals(baseTxnId)) {
+      myBaseTxnCommitInfo.reset();
+      myBaseTxnCommitInfo.update(null, baseTxnId, 0);
+      sendProjectStateLogs();
```
dbarashev, line 1023: remove it? Sending logs doesn't make too much sense if we just have reset the state.

npostnikova: In such a case, we'll send all the logs from the beginning, one by one
Where should we start sending logs instead? I thought that it's the right place

ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
npostnikova: Hmmm
We perform websocket handshake (with base txn id & project refid exchange) only under this check. 
If we remove the check, we'll log an error every time a new log record is added for an offline document, won't we?

ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
dbarashev: Send the logs when an undoable edit happens. But I don't think that a user has any chance to make an undoable edit immediately at the moment when we reset the log.

ganttproject/src/main/java/net/sourceforge/ganttproject/GanttProject.java
dbarashev: Well, in this case we shall check "if we are working with an online document, in any sense", not "is it a local online document setup". 

We probably need to modify the local dev server so that it could respond to project read and write operations, just like a regular GP Cloud server. In this case the check may look like `if (project.document is OnlineDocument)`

Issue #2172, opened 2022-06-26
Refactor the procedure of establishing colloboque connection


dbarashev:
We need to integrate the code which initializes colloboque stuff on the client and on the server side with the cloud document open action.

PR #2173, opened 2022-06-26
Refactored Colloboque client initialization

dbarashev:
- some colloboque code from GanttProject class extracted into ColloboqueClient.kt
- onboard() call moved from the StorageDialogBuilder.kt into ProjectUIFacadeImpl.kt, It  will be called not only when document is opened from the UI, but when it is opened from the command line too
- initialization of the ColloboqueClient integrated with GPCloudDocument contents reading and onboarding

Update #2172


dbarashev:
@npostnikova this is FYI if you plan to touch code around this. Now communication between the Desktop and Cloud/Colloboque is close to what will be in the reality.

PR #2193, opened 2022-07-11
An attempt to generate SQL statements from json-like dto on Colloboque side

edgarzhavoronkov:
Client side is in progress, but a review is highly appreciated

### DIFF
--- a/ganttproject/src/main/java/net/sourceforge/ganttproject/storage/Xlog.kt
+++ b/ganttproject/src/main/java/net/sourceforge/ganttproject/storage/Xlog.kt
@@ -29,6 +29,40 @@ data class InitRecord(
   val projectXml: String
 )
 
+@Serializable
+sealed class OperationDto() {
+  @Serializable
+  data class InsertOperationDto(
+    val tableName: String,
+    val values: Map<String, String>
+  ): OperationDto()
+
+  @Serializable
+  data class UpdateOperationDto(
+    val tableName: String,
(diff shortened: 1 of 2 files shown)


ganttproject/src/main/java/net/sourceforge/ganttproject/storage/Xlog.kt
```diff
@@ -29,6 +29,40 @@ data class InitRecord(
...
+@Serializable
+sealed class OperationDto() {
+  @Serializable
+  data class InsertOperationDto(
+    val tableName: String,
+    val values: Map<String, String>
+  ): OperationDto()
+
+  @Serializable
+  data class UpdateOperationDto(
```
dbarashev, line 41: I think we need a condition for update as well. We use task uid for that.

dbarashev: Add the processing of these conditions in ColloboqueServer::applyOperation ?

edgarzhavoronkov: Yup, forgot to include file in commit

dbarashev (approved):
LGTM
I think we need some tests to check that json translates to some expected SQL

edgarzhavoronkov:
Ok, i finished generating and sending it on the client side, some debugging and i'll add tests then

PR #2238, opened 2022-09-14
An attempt to apply log from Colloboque on client

edgarzhavoronkov:
Not sure if it is a good idea to remove the original task, but nevertheless, a starting point for polishing


ganttproject/src/main/java/net/sourceforge/ganttproject/task/TaskManagerImpl.java
```diff
@@ -1032,4 +1039,49 @@ public Boolean isZeroMilestones() {
...
+        taskImpl.setNotes(record.getNotes());
+        if (record.getIsMilestone()) {
+          taskImpl.setMilestone(record.getIsMilestone());
+        }
+
+        Task originalTask = getTask(record.getNum());
+        myTaskMap.myId2task.put(record.getNum(), taskImpl);
+        myDependencyGraph.addTask(taskImpl);
+
+        Task prevSibling = getTaskHierarchy().getPreviousSibling(originalTask);
```
dbarashev, line 1068: We don't have any information about the siblings/parents in the database itself, do we? 

I think we want such information to be in the database, because this code will definitely break, should we start doing anything more complex than task property values change. Perhaps, we could use the outline number for this purpose.

This is for the subsequent pull requests though. Meanwhile, I suggest the following:

1. build the map `task id => (parent id, position)` from the "original" hierarchy, as we have no other, before we run this method. 
2. pass the map as the argument to this method. 
3. run this code in a buffer project or empty task manager, like we do in importers
4. then call importData to import the result

It will make this code and the whole process less fragile and easier to maintain

edgarzhavoronkov: Hm, done, but it still keeps original task even if i try to delete it somewhy(i probably missing something)

PR #2280, opened 2022-11-18
Bump kotlinx-serialization to the same version as in the GanttProject

edgarzhavoronkov:
I noticed that it fixes NoClassDefFoundError in the runtime of Colloboque


dbarashev:
Yup, thanks!

Issue #2291, opened 2022-12-06
Accidental file overwrite using "Recent Documents"


ZaneUJi:
The version is 3.2.3230.

Here is the steps to reproduce the problem.
1. Create a new project, save it as a.gan
2. Create another project
3. Press Ctrl + S or click the save icon in the toolbar to open the file name dialog
4. Select "Recent Documents" in the left area, then select a.gan. It should look like below:
![image](https://user-images.githubusercontent.com/120994/205827839-6e418d78-c80f-498d-9913-6d636378efd0.png) 
5. Imput a new name, i.e. change a.gan to something like aaa.gan to make sure that the recent file list is empty.
6. Click Save button.
7. Then you'll find that a.gan has been overwritten.

I understand that some people don't like the comfirmation dialog. But it doesn't seem right to apply the user input name to clear the file list then ignore this name and overwrit the original file.

PR #2326, opened 2023-07-11
Make accidental overwriting less probable

dbarashev:
- Disable file name text field in SAVE mode
- Require "overwrite" checkbox to be checked when using double-click


dbarashev:
Update #2291

Issue #2450, opened 2024-04-25
Validation in the date range option in PDF export 


dbarashev:
PDF export allows for choosing a date range for the chart image export. It is possible to define a range where the start date is after the end date. Apparently, this won't affect PDF export (there is only a task table image with no chart in the PDF file) but leads to crashes in the HTML export:

```
Caused by: java.lang.IllegalArgumentException: Width (0) and height (793) must be > 0
	at java.desktop/java.awt.image.SampleModel.<init>(SampleModel.java:127)
	at java.desktop/java.awt.image.SinglePixelPackedSampleModel.<init>(SinglePixelPackedSampleModel.java:144)
	at java.desktop/java.awt.image.SinglePixelPackedSampleModel.<init>(SinglePixelPackedSampleModel.java:110)
	at java.desktop/java.awt.image.PackedColorModel.createCompatibleSampleModel(PackedColorModel.java:312)
	at net.sourceforge.ganttproject.chart.export.RenderedChartImage.<init>(RenderedChartImage.java:51)
	at net.sourceforge.ganttproject.chart.ChartRasterImageBuilder.acceptChart(PrintChartApiImpl.kt:103)
	at net.sourceforge.ganttproject.chart.export.ChartImageBuilder.buildImage(ChartImageBuilder.kt:56)
	at net.sourceforge.ganttproject.chart.PrintChartApiImpl.buildImage(PrintChartApiImpl.kt:55)
	at net.sourceforge.ganttproject.chart.PrintChartApiImpl.exportChart(PrintChartApiImpl.kt:49)
	at org.ganttproject.impex.htmlpdf.ExporterToHTML$1.run(ExporterToHTML.java:87)
	at net.sourceforge.ganttproject.export.ExporterBase$2.run(ExporterBase.java:180)
	... 2 more
 ```

PR #2451, opened 2024-04-29
Validate date range during export to PDF/HTML

dbarashev:
- added export date range options to HTML export
- added date range validation
- migrated from Apache Pair to kotlin Pair


dbarashev:
Update ussue #2450

PR #2795, opened 2026-04-16
Fix NPE on startup: DEFAULT_TRANSLATION_LOCALIZER stays DummyLocalizer when locale file exists but is missing keys

### LINKED ISSUE #2794: NPE on startup: DEFAULT_TRANSLATION_LOCALIZER stays DummyLocalizer when locale has its own properties file
## Summary

GanttProject fails to start with a `NullPointerException` when the user's language (either the JVM default or the `ui.language` setting in `.ganttproject`) resolves to a locale that has its own `i18n_xx_YY.properties` file, **but** that file is missing keys that exist only in the base `i18n.properties`.

The crash manifests as:
```
java.lang.NullPointerException: Cannot invoke "String.length()" because "pattern" is null
    at java.text.MessageFormat.applyPattern(MessageFormat.java:469)
    at java.text.MessageFormat.<init>(MessageFormat.java:383)
    at java.text.MessageFormat.format(MessageFormat.java:883)
    at net.sourceforge.ganttproject.action.ViewToggleAction.getLocalizedDescription(ViewToggleAction.java:44)
    at net.sourceforge.ganttproject.action.GPAction.updateTooltip(GPAction.java:270)
    at net.sourceforge.ganttproject.action.GPAction.<init>(GPAction.java:113)
    ...
    at biz.ganttproject.app.GanttProjectFxApp.start(GanttProjectFxApp.kt:59)
```

## Affected key

`view.toggle.description` exists in `i18n.properties` (and in most non-English locale files such as `i18n_it_IT.properties`) but is **absent** from several locale files including:
- `i18n_en_GB.properties`
- `i18n_en_US.properties`
- `i18n_en_AU.properties`
- `i18n_en.properties`

Any user whose `ui.language` resolves to one of these locales will hit the crash on startup.

## Root cause

The issue is in the interaction between `DEFAULT_TRANSLATION_LOCALIZER` and the `defaultTranslation` lazy initializer.

**`InternationalizationCore.kt`**, line 109:
```kotlin
var DEFAULT_TRANSLATION_LOCALIZER: Localizer = DummyLocalizer
```

`DEFAULT_TRANSLATION_LOCALIZER` starts as `DummyLocalizer` (which always returns `null`). It is reassigned to a real `SingleTranslationLocalizer` only as a **side-effect** of accessing the `defaultTranslation` lazy val:

**`InternationalizationImpl.kt`**, lines 80–84:
```kotlin
val defaultTranslation by lazy {
  (_explicitTranslation ?: createTranslationFromFile(Locale.ENGLISH, "i18n.properties") ?: Translation(Locale.ENGLISH) { null }).also {
    DEFAULT_TRANSLATION_LOCALIZER = SingleTranslationLocalizer(it)  // <-- only executed on first access of defaultTranslation
  }
}
```

**`Internationalization.kt`**, `setLocale`:
```kotlin
fun setLocale(locale: Locale) {
  ourLocale = locale
  ourCurrentTranslation.value = createTranslation(locale) ?: defaultTranslation
}
```

If `createTranslation(locale)` returns a non-null value (which it does whenever `i18n_xx_YY.properties` exists for that locale), the `?: defaultTranslation` short-circuit never fires, so `defaultTranslation` is **never accessed** and `DEFAULT_TRANSLATION_LOCALIZER` remains `DummyLocalizer`.

Later, in `DefaultLocalizer.formatTextOrNull`:
```kotlin
tr.mapKey(prefixedKey)?.let { value ->
    MessageFormat.format(value, *args)
} ?: this.baseLocalizer().formatTextOrNull(key, *args)  // baseLocalizer() returns DEFAULT_TRANSLATION_LOCALIZER
```

When the current locale's file is missing a key, `tr.mapKey(prefixedKey)` returns `null`, and the fallback `baseLocalizer()` (i.e. `DEFAULT_TRANSLATION_LOCALIZER`) is consulted — but it is still `DummyLocalizer`, so it returns `null` too. The caller (`ViewToggleAction.getLocalizedDescription`) then passes `null` as the pattern to `MessageFormat.format`, causing the NPE.

## Proposed fix

Force `defaultTranslation` to be initialized (which sets `DEFAULT_TRANSLATION_LOCALIZER`) **unconditionally** inside `setLocale`, regardless of whether a locale-specific translation exists:

**`ganttproject/src/main/java/biz/ganttproject/app/Internationalization.kt`**:
```kotlin
fun setLocale(locale: Locale) {
  ourLocale = locale
  // Touching defaultTranslation ensures that DEFAULT_TRANSLATION_LOCALIZER is initialized
  // (its assignment is a side-effect of the lazy initializer). Without it, locales whose
  // properties file is missing keys fall back to a DummyLocalizer and return null.
  val fallback = defaultTranslation
  ourCurrentTranslation.value = createTranslation(locale) ?: fallback
}
```

This single-line change guarantees that `DEFAULT_TRANSLATION_LOCALIZER` is always a proper `SingleTranslationLocalizer` backed by `i18n.properties` before any key lookup happens.

### Verified

With this fix applied, GanttProject starts successfully with `user.language=it` (and with the project settings `ui.language=en_GB` that triggered the original report), reaching `ProjectOpenActivityCompleted` without any exception.

## Environment

- Java: Eclipse Adoptium Temurin 21.0.10+7-LTS (but the bug is not JDK-specific)
- OS: Windows 10 (same logic applies cross-platform)
- GanttProject version: 3.4.3392 (built from `master` / recent HEAD)

## Reproduction

1. Set `ui.language` to `en_GB` in `~/.ganttproject` (or use any locale whose `i18n_xx_YY.properties` is missing `view.toggle.description`)
2. Launch GanttProject
3. App fails immediately with the NPE shown above

## Notes

- The bug is latent: it fires whenever a key present in `i18n.properties` is missing from a locale-specific file *and* that locale file exists (so `createTranslation` returns non-null). Adding `view.toggle.description` to the missing locale files would mask the symptom but not fix the underlying issue — any future key added to `i18n.properties` without being propagated to all locale files would re-trigger it.
- A more robust long-term fix might be to initialize `DEFAULT_TRANSLATION_LOCALIZER` eagerly at module startup rather than as a lazy side-effect.

mspinolaeie:
## Problem

GanttProject crashes on startup with:
```
java.lang.NullPointerException: Cannot invoke "String.length()" because "pattern" is null
    at java.text.MessageFormat.applyPattern(MessageFormat.java:469)
    at net.sourceforge.ganttproject.action.ViewToggleAction.getLocalizedDescription(ViewToggleAction.java:44)
    at net.sourceforge.ganttproject.action.GPAction.updateTooltip(GPAction.java:270)
    at biz.ganttproject.app.GanttProjectFxApp.start(GanttProjectFxApp.kt:59)
```

Any user whose UI language resolves to a locale that has its own `i18n_xx_YY.properties` file, but where that file is **missing** keys present only in the base `i18n.properties`, will hit this. Confirmed affected: `en_GB`, `en_US`, `en_AU`, `en`. The key `view.toggle.description` exists in `i18n.properties` and most non-English locale files, but not in any of the English-variant files.

Closes #2794

## Root cause

`DEFAULT_TRANSLATION_LOCALIZER` (in `InternationalizationCore.kt`) starts as `DummyLocalizer` and is only reassigned inside the `defaultTranslation` lazy initializer:

```kotlin
// InternationalizationImpl.kt
val defaultTranslation by lazy {
  (...).also {
    DEFAULT_TRANSLATION_LOCALIZER = SingleTranslationLocalizer(it)  // only runs on first access
  }
}
```

In `setLocale()`:
```kotlin
ourCurrentTranslation.value = createTranslation(locale) ?: defaultTranslation
```

When `createTranslation(locale)` succeeds (i.e. the locale file exists), the `?: defaultTranslation` branch is never evaluated, so `DEFAULT_TRANSLATION_LOCALIZER` stays as `DummyLocalizer`.

In `DefaultLocalizer.formatTextOrNull`, when a key is missing from the current locale's file, control falls through to `baseLocalizer()` which returns `DEFAULT_TRANSLATION_LOCALIZER` — still a `DummyLocalizer` — giving `null`. `ViewToggleAction.getLocalizedDescription` then passes that `null` as pattern to `MessageFormat.format`, causing the NPE.

## Fix

Evaluate `defaultTranslation` unconditionally in `setLocale()` to guarantee `DEFAULT_TRANSLATION_LOCALIZER` is initialized before any key lookup:

```kotlin
fun setLocale(locale: Locale) {
  ourLocale = locale
  // Touching defaultTranslation ensures that DEFAULT_TRANSLATION_LOCALIZER is initialized
  // (its assignment is a side-effect of the lazy initializer). Without it, locales whose
  // properties file is missing keys fall back to a DummyLocalizer and return null.
  val fallback = defaultTranslation
  ourCurrentTranslation.value = createTranslation(locale) ?: fallback
}
```

## Testing

Verified with `user.language=it`, `user.country=IT` and UI settings `ui.language=en_GB` (the original failure scenario): GanttProject now starts successfully and reaches `ProjectOpenActivityCompleted` without any exception.

Note: a more robust long-term approach would be to initialize `DEFAULT_TRANSLATION_LOCALIZER` eagerly at module load time rather than as a lazy side-effect, to avoid similar fragility in the future.

### DIFF
--- a/ganttproject/src/main/java/biz/ganttproject/app/Internationalization.kt
+++ b/ganttproject/src/main/java/biz/ganttproject/app/Internationalization.kt
@@ -29,7 +29,11 @@ private var ourLocale: Locale = Locale.getDefault()
 
 fun setLocale(locale: Locale) {
   ourLocale = locale
-  ourCurrentTranslation.value = createTranslation(locale) ?: defaultTranslation
+  // Touching defaultTranslation ensures that DEFAULT_TRANSLATION_LOCALIZER is initialized
+  // (its assignment is a side-effect of the lazy initializer). Without it, locales whose
+  // properties file is missing keys fall back to a DummyLocalizer and return null.
+  val fallback = defaultTranslation
+  ourCurrentTranslation.value = createTranslation(locale) ?: fallback
 }
 
 fun getCurrentLocale() = ourLocale


dbarashev:
Thanks for the investigation and the fix!

Issue #2312, opened 2023-06-12
Boolean flag indicating whether a task is on a critical path


dbarashev:
It makes sense to export information about critical tasks to CSV/Excel and make it available in the calculated columns. Read-only boolean flag `is_critical` may help to those who analyzes project data with external tools.

PR #2799, opened 2026-04-18
Immediately display "Is Critical" checkboxes when the critical path is switched on.

dbarashev:
Update issue #2312


dbarashev:
@copilot please review

PR #2806, opened 2026-06-22
Introduce 'Document Ready' state in project open process

dbarashev:
```
### Summary

This pull request integrates more project opening logic into the Project-Open state machine. Three new states allow for waiting for "Document Ready" state and for handling document fork and authentication needed states.

Besides important bugfixes: 
 - a websocket would not open if we just open a cloud document
- chart would freeze when opening a cloud document (related to the Decorator implementation)


dbarashev:
@claude review

PR #2821, opened 2026-08-21
Fix GPCloudDocumentTest failing on Windows for absolute mirror paths

Natalie-the-technician:
### What fails

`GPCloudDocumentTest.testMirroredAutomaticallyWhenHasOptions` fails on Windows. On Linux the same test passes, so the problem is invisible on a Linux-only run.

```
GPCloudDocumentTest > testMirroredAutomaticallyWhenHasOptions() FAILED
    java.io.IOException: Failed to create parent directories to file
    C:\Users\<user>\AppData\Local\Temp\1787299014323-0\C:\Users\<user>\AppData\Local\Temp\1787299014323-0\727eacecd5b09a72
        at net.sourceforge.ganttproject.document.FileDocument.create(FileDocument.kt:132)
        at biz.ganttproject.storage.cloud.GPCloudDocument.saveOfflineMirror(GPCloudDocument.kt:266)
        at biz.ganttproject.storage.cloud.GPCloudDocument.getInputStream(GPCloudDocument.kt:306)
        at biz.ganttproject.storage.cloud.GPCloudDocumentTest.testMirroredAutomaticallyWhenHasOptions(GPCloudDocumentTest.kt:210)
```

The temporary directory appears twice in the path, drive letter and all, and `mkdirs()` cannot create it.

### Why it happens

The test stores an **absolute** path in the cloud options (`it.offlineMirror = mirrorFile.absolutePath`) and later passes that same path through `File(testMirrorFolder, path)`. Per the javadoc of that constructor, an absolute child is converted "into a relative pathname in a system-dependent way". On Linux the result is a legal path; on Windows it becomes a path containing two drive letters.

### What this changes

A small helper, `resolveMirrorFile(path)`, returns the path unchanged when it is already absolute and only joins it with the test mirror folder otherwise. The two call sites of `File(testMirrorFolder, path)` in the offline document factory now use it.

The name avoids colliding with the existing local variable `mirrorFile` in the same class.

### Note on the second call site

The second replacement is inside `testOfflineGoesOnlineOnReconnect`, which carries no `@Test` annotation and therefore does not run (the comment above it says it is currently ignored). That change is there for consistency, not because it fixes an observed failure.

### Verified

| | tests | failures |
|---|---|---|
| Windows, before | 10 | 1 |
| Windows, after | 10 | 0 |
| Linux, before | 10 | 0 |
| Linux, after | 10 | 0 |

The Linux rows are there to show the change breaks nothing — on Linux the join produces a legal path, so that platform cannot distinguish the two states either way.

### DIFF
--- a/ganttproject/src/test/java/biz/ganttproject/storage/cloud/GPCloudDocumentTest.kt
+++ b/ganttproject/src/test/java/biz/ganttproject/storage/cloud/GPCloudDocumentTest.kt
@@ -82,7 +82,7 @@ class GPCloudDocumentTest {
 
   private fun prepareReadCall(doc: GPCloudDocument, responseBuilder: ()->TestResponse) {
     doc.httpClientFactory = { mockHttpClient }
-    doc.offlineDocumentFactory = { path -> FileDocument(File(testMirrorFolder, path)) }
+    doc.offlineDocumentFactory = { path -> FileDocument(testMirrorFolder.resolve(path)) }
     doc.executor = MoreExecutors.newDirectExecutorService()
 
     EasyMock.reset(mockHttpClient)
@@ -282,7 +282,7 @@ class GPCloudDocumentTest {
     val executor = Executors.newSingleThreadExecutor()
     val doc = GPCloudDocument(teamRefid = "team1", teamName = "Team 1", projectRefid = "prj1", projectName = "Project 1", projectJson = null)
     doc.httpClientFactory = { mockHttpClient }
-    doc.offlineDocumentFactory = { path -> FileDocument(File(testMirrorFolder, path)) }
+    doc.offlineDocumentFactory = { path -> FileDocument(testMirrorFolder.resolve(path)) }
     doc.executor = executor
     doc.isNetworkAvailable = Callable {
       if (retryCounter.count > 0) {


ganttproject/src/test/java/biz/ganttproject/storage/cloud/GPCloudDocumentTest.kt
```diff
@@ -80,9 +80,21 @@ class GPCloudDocumentTest {
...
+  /**
+   * Joins [path] with the test mirror folder, unless [path] is already absolute.
+   *
+   * `File(parent, child)` converts an absolute child "into a relative pathname in a
+   * system-dependent way" (javadoc of that constructor). On Linux the result is a legal path, on
+   * Windows it is a path with two drive letters that cannot be created. The mirror location kept
+   * in the cloud options is an absolute path, so joining it unconditionally makes these tests
+   * pass or fail depending on the platform.
+   */
+  private fun resolveMirrorFile(path: String): File =
```
dbarashev: I think that [resolve extension function](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.io/resolve.html) in Kotlin libs will do the same

Natalie-the-technician: Good catch. `File.resolve(String)` does exactly that: it returns the argument unchanged when it is absolute and joins otherwise, so the helper and its comment both disappear.

I'll push an update replacing `resolveMirrorFile(path)` with `testMirrorFolder.resolve(path)` and re-run the test on Windows to confirm it still distinguishes the two states.

Natalie-the-technician: Done — replaced with `testMirrorFolder.resolve(path)`. The diff is down to the two call sites.

Verified on Windows on the same build: without the change the test still fails with the doubled path, with it the class passes 10 of 10. So it still measures something after the switch.

PR #2822, opened 2026-08-21
Exit the application when quitting without saving

Natalie-the-technician:
### What fails

Modify a project, choose **Project → Exit**, then **Don't save**. The window stays open and the process keeps running.

Measured on the same machine, same steps:

| | process | window |
|---|---|---|
| before | PID alive, 576 MB, 65 threads | `* GanttProject[…]`, responsive |
| after | no java process | gone |

The last log line in both cases is `[GanttOptions] save(): finished!!` — `quitApplication()` completes correctly, saves the options and closes the project. Nothing then terminates the process.

### Why it happens

`MainApplication` sets `myLock` in the quit callback and calls `notify()`. The block that *waits* on it is commented out — from the commit "commented out or removed usages of the main application window". With no observer, `launch()` returns immediately, and the `System.exit(0)` further down is reached while `myLock` is still false.

### What this changes

The callback terminates directly when `withSystemExit` is true.

The condition matters: `withSystemExit` is not always true. Two call sites pass false, both in the updater — `PlatformOptionPage.kt:119` and `Update.kt:47` — where the intent is to restart rather than quit. The added block is skipped on that path, so restart behaviour is unchanged.

### What is not covered

I verified the updater path by reading it end to end (`quitApplication(false)` → `doQuitApplication` → `ourQuitCallback.accept(false)` → the condition is false, no exit), but I did not trigger it: that needs a real update available from the update service.

No test touches `MainApplication` — the change is covered by the manual before/after above, not by an automated test.

### DIFF
--- a/ganttproject/src/main/java/net/sourceforge/ganttproject/application/MainApplication.java
+++ b/ganttproject/src/main/java/net/sourceforge/ganttproject/application/MainApplication.java
@@ -100,10 +100,22 @@ public class MainApplication implements IPlatformRunnable {
     }
 
 
+    // The block below that waits on myLock is commented out, so nothing observes the notify()
+    // and launch() returns immediately. By the time the System.exit(0) further down is reached,
+    // myLock is still false and the process keeps running with its window open, even though
+    // quitApplication() has already saved the options and closed the project. Terminate here
+    // instead.
+    //
+    // withSystemExit is not always true: the updater passes false to restart rather than quit.
     Consumer<Boolean> onApplicationQuit = withSystemExit -> {
       synchronized(myLock) {
         myLock.set(withSystemExit);
         myLock.notify();
+      }
+      if (withSystemExit) {
+        logger.debug("Application quit requested, terminating");
+        GPLogger.close();
+        System.exit(0);
       }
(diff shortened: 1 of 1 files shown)


dbarashev:
@Natalie-the-technician the lock is not needed anymore, because `appBuilder.launch()` call is blocking (there is `job.join()` inside App.kt), so the fix is to just call Platform.exit in the ExitAction

Natalie-the-technician:
Thanks for taking it — your fix is in the better place. I was looking at the callback and missed that `ExitAction` never calls `Platform.exit()` at all.

One observation that might be worth a moment while you are in there. I measured that nothing observes the `notify()` on `myLock`: the block that waits on it is commented out, from the commit "commented out or removed usages of the main application window". If `launch()` blocks anyway, then `myLock` and its callback are not merely redundant, they are dead — and reasoning about them is what led me to the wrong place. Removing them would save the next person the same detour.

Happy to close this PR in favour of yours. Just say the word, or close it yourself if that is easier.

dbarashev:
There is no more waiting on myLock, it only carries the system exit flag.

PR #2823, opened 2026-08-21
fix: Task Properties dialog: raw i18n keys and $Resources tab title

### LINKED ISSUE #2800: Task Properties dialog: raw i18n keys and $Resources tab title
## Summary

Two related i18n issues in the **Task Properties** dialog introduced with the new JavaFX properties panel.

---

## Bug 1 — Field labels shown as raw translation keys

**File:** `ganttproject/src/main/java/net/sourceforge/ganttproject/gui/taskproperties/MainPropertiesPanel.kt`

All field labels in the General tab display as raw keys (`name.label`, `startDate.label`, `progress.label`, etc.) instead of the localized strings ("Name", "Begin date", "Progress", …).

### Root cause

The `private val i18n` at the bottom of the file builds a 4-step fallback chain, but `master` only has 3 steps:

| Step | Looks up | Result for e.g. `name.label` |
|------|----------|-------------------------------|
| 1 | `option.taskProperties.main.name.label` in current locale | ❌ key not in translation file |
| 2 | `name` (legacy transform) in current locale | ❌ may not resolve |
| 3 | `option.taskProperties.main.name.label` in `RootLocalizer` | ❌ key not in file |
| **4 (missing)** | **`name` (legacy transform) in `RootLocalizer`** | **✅ "Name" found** |

Step 4 is absent. Because `RootLocalizer` is a `DefaultLocalizer` backed by the Eclipsito translation registry (which loads `i18n.properties` from the classpath), `RootLocalizer.formatTextOrNull("name")` always resolves to `"Name"` (English). Without step 4, field labels fall through all steps unresolved and are rendered as-is.

Note: section titles (`section.main`, `section.view`, `section.documents`) already have structured keys in `i18n.properties`, so they resolve at step 1 and are unaffected.

### Fix

Add the missing fallback step to the chain in `MainPropertiesPanel.kt`:

```kotlin
private val i18n = i18n {
  default(withFallback = false)
  prefix("option.taskProperties.main") {
    default(withFallback = false)
    transform(legacyKeyTransform)
    fallback {
      default()
      prefix("option.taskProperties.main")
      fallback {
        // NEW: fall back to English legacy key via RootLocalizer
        default()
        transform(legacyKeyTransform)
      }
    }
  }
}
```

---

## Bug 2 — Resources tab title shows `$Resources`

**File:** `ganttproject/src/main/java/net/sourceforge/ganttproject/gui/taskproperties/TaskResourcesPanel.kt`

The **Resources** tab in the Task Properties dialog shows `$Resources` (or `$Risorse` in Italian) instead of the plain string.

### Root cause

```kotlin
// line 105 — master
val title: String = i18n.formatText("human")
```

The translation key `human` has value `$Resources` (EN) / `$Risorse` (IT) — the `$` is a legacy Swing mnemonic marker, historically stripped by `GanttLanguage.correctLabel()`. The new JavaFX code calls `RootLocalizer.formatText()` directly, bypassing that stripping.

### Fix

```kotlin
val title: String = i18n.formatText("human").removeMnemonicsPlaceholder()
```

`removeMnemonicsPlaceholder()` is already defined in `Internationalization.kt`.

---

Roshan931:
Fixes #2800

## Root cause

Resources tab title not stripped of legacy Swing mnemonic marker

## Changes

- Apply removeMnemonicsPlaceholder() to the localized "human" string used as the Resources tab title in TaskResourcesPanel.kt. Bug 1 (raw i18n keys in MainPropertiesPanel.kt) is already fixed in the checked-out revision by commit 47f1430c2, so no change was made there.

### DIFF
--- a/ganttproject/src/main/java/net/sourceforge/ganttproject/gui/taskproperties/TaskResourcesPanel.kt
+++ b/ganttproject/src/main/java/net/sourceforge/ganttproject/gui/taskproperties/TaskResourcesPanel.kt
@@ -22,6 +22,7 @@ import biz.ganttproject.app.FXThread
 import biz.ganttproject.app.PropertyPane
 import biz.ganttproject.app.PropertyPaneBuilderImpl
 import biz.ganttproject.app.RootLocalizer
+import biz.ganttproject.app.removeMnemonicsPlaceholder
 import biz.ganttproject.core.option.ObservableBoolean
 import biz.ganttproject.core.option.ObservableMoney
 import javafx.beans.property.SimpleBooleanProperty
@@ -102,7 +103,7 @@ class TaskResourcesPanel(
   private val costIsCalculated = ObservableBoolean("option.taskProperties.cost.calculated.label", task.cost.isCalculated)
   private val costValue = ObservableMoney("option.taskProperties.cost.value", task.cost.value)
 
-  val title: String = i18n.formatText("human")
+  val title: String = i18n.formatText("human").removeMnemonicsPlaceholder()
   private val tableAndActions = AbstractTableAndActionsComponentFx(tableView, model)
 
   val fxComponent by lazy {


dbarashev:
This doesn't actually change anything. The issue #2800 is not reproducible in the latest `master` branch, because we already remove placeholders inside Internationalization infrastructure code.

PR #2824, opened 2026-08-21
fix: Logo is not scaled down in PDF export

### LINKED ISSUE #2613: Logo is not scaled down in PDF export
Scenario:

1. Open any project
2. In the settings, choose a logo image with significant height, e.g. > 300px
3. Export the project to PDF

Expected: in the PDF export, the logo is scaled down, just like in the application window
Actual: the logo is not scaled down. Instead, the timeline height grows.

Roshan931:
Fixes #2613

## Root cause

Export path draws the project logo at native size instead of scaling it

## Changes

- ChartDimensions now caps the logo height at the default logo height (47px, same height the app window renders the logo at) and exposes an aspect-ratio-preserving logoWidth; both export visitors (iText PDF ChartWriter and the raster PNG/print builder) draw the logo scaled into that box and tolerate a null logo. Added a unit test for the new dimension arithmetic.

### DIFF
--- a/ganttproject/src/main/java/net/sourceforge/ganttproject/chart/export/ChartDimensions.kt
+++ b/ganttproject/src/main/java/net/sourceforge/ganttproject/chart/export/ChartDimensions.kt
@@ -19,9 +19,19 @@ along with GanttProject.  If not, see <http://www.gnu.org/licenses/>.
 package net.sourceforge.ganttproject.chart.export
 
 import net.sourceforge.ganttproject.GanttExportSettings
+import net.sourceforge.ganttproject.gui.UIFacade
+import kotlin.math.min
+import kotlin.math.roundToInt
 
 class ChartDimensions internal constructor(settings: GanttExportSettings, treeTable: TreeTableApi) {
-  val logoHeight: Int = settings.logo.getHeight(null)
+  private val logoSourceWidth: Int = settings.logo?.getWidth(null) ?: -1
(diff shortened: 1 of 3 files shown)


ganttproject/src/main/java/net/sourceforge/ganttproject/chart/export/ChartDimensions.kt
```diff
@@ -19,9 +19,19 @@ along with GanttProject.  If not, see <http://www.gnu.org/licenses/>.
 package net.sourceforge.ganttproject.chart.export
 
 import net.sourceforge.ganttproject.GanttExportSettings
+import net.sourceforge.ganttproject.gui.UIFacade
+import kotlin.math.min
+import kotlin.math.roundToInt
 
 class ChartDimensions internal constructor(settings: GanttExportSettings, treeTable: TreeTableApi) {
-  val logoHeight: Int = settings.logo.getHeight(null)
+  private val logoSourceWidth: Int = settings.logo?.getWidth(null) ?: -1
```
dbarashev, line 27: I suggest passing the logo image to the constructor instead of accessing `DEFAULT_LOGO` directly, and using local variables in the constructor instead of introducing private fields.

dbarashev (approved):
Thanks for contributing! I left a few non-blocking comments below, but I'm going ahead and merging this now. I’ll follow up with a few quick tweaks upstream—feel free to check those out if you plan to contribute again!

PR #2830, opened 2026-08-31
Notify listeners when days off are removed, and add the missing removal methods

Natalie-the-technician:
### What is wrong

`HumanResource.addDaysOff()` has no counterpart. The only way to take an absence
away is to reach into the `DefaultListModel` that `getDaysOff()` hands out and
mutate it directly, and nothing observes that model. Adding an absence and
removing one are therefore not the same kind of operation: adding resets the
cached `LoadDistribution` and fires `fireResourceChanged()`, removing does
neither.

### Steps to reproduce

1. Create a resource and give it exactly one day off. Press OK.
2. Open the Resources Chart. The absence is drawn as a yellow block.
3. Open the resource again, delete the day off, press OK.

Expected: the block disappears, and anything listening to the resource model is
told that something changed.

Actual: the block stays, and no `resourceChanged` event is fired at all.

### Cause

`GanttDialogPerson.applyChanges()` clears the list before writing the edited
intervals back:

```java
person.getDaysOff().clear();
for (DateInterval interval : myDaysOffModel.getIntervals()) {
  person.addDaysOff(new GanttDaysOff(interval.getStart(), interval.getEnd()));
}
```

With at least one interval left, the last `addDaysOff()` repairs both by
accident. Delete the *last* absence and the loop body never runs, so nothing
calls `resetLoads()` and nothing fires. `resetLoads()` drops the memoised
`myLoadDistribution`, which is the only cache of the computed absences — that is
why the chart keeps drawing the old picture.

That dialog line is the symptom rather than the defect. A fix that only makes
that one line fire leaves the getter handing out an unobserved mutable model,
and the next caller that removes an absence — a batch edit, an importer, a
script — reintroduces the same gap.

### What this changes

**1. `getDaysOff()` hands out an unmodifiable `List` view, and every change goes
through a method.** The field is a plain `ArrayList<GanttDaysOff>` and the getter
returns `Collections.unmodifiableList` over it, typed `List<GanttDaysOff>` —
`DefaultListModel` is a Swing class and models a UI control, not a resource. A
view rather than a copy, so a caller holding on to it keeps seeing the resource.

`addDaysOff()`, `removeDaysOff()` and `clearDaysOff()` each call
`onDaysOffChanged()`, which does `resetLoads()` and `fireResourceChanged()`.
The two halves depend on each other: a plain `List` cannot be observed, so the
notification has to live in the methods, and that is only sound because no
caller can reach past them any more.

`removeDaysOff()` fires only when something was really removed, and
`clearDaysOff()` returns early on an already empty list — what
`DefaultListModel.removeElement()` and `clear()` did by themselves.

The copying constructor fills the list directly with `addAll()` rather than
going through `addDaysOff()`, so copying stays silent without depending on
`areEventsEnabled` for the days off at all.

**2. The two missing counterparts**, in the shape `HumanResourceManager` already
uses for its own list (add / remove / clear):

```java
boolean removeDaysOff(GanttDaysOff gdo)
void clearDaysOff()
```

`removeDaysOff` returns whether anything was removed, which is also the answer to
"did the listeners hear about it" — nothing fires when the interval was not
there. It is `clearDaysOff` rather than a bare `clear()` because a resource holds
assignments and custom properties too.

One caveat, documented on the method and pinned by a test rather than changed
here: `removeDaysOff` matches by `Object.equals`, the way the list matches.
`GanttDaysOff` only overloads `equals(GanttDaysOff)` and does not override
`equals(Object)`, so an interval built afresh from the same two dates is not the
one the resource holds. Callers should pass an instance obtained from the
resource.

`GanttDialogPerson.applyChanges()` now calls `clearDaysOff()` instead of
`getDaysOff().clear()`, and no production code can write to the handed-out list
any more. The readers of `getDaysOff()` move from `getSize()`/`getElementAt(i)`
to `size()`/`get(i)`; `LoadDistribution` and `ProjectFileExporter` were reading
through a raw `DefaultListModel` and lose their casts.

### The number of notifications does not change

Clearing and then writing N intervals back still costs `(M > 0 ? 1 : 0) + N`
`resourceChanged` events, where M is the number of intervals before OK: one event
for the whole removal, none at all when there was nothing to remove. With a plain
`List` that is not free — `List.clear()` has no notion of firing — so
`clearDaysOff()` returns early on an empty list and `removeDaysOff()` returns
`false` without firing when nothing matched. The single event this PR adds is
exactly the one that was missing in the "delete the last absence" case.

### Tests

`HumanResourceDaysOffTest`

* `removing the last day off notifies the listeners just as adding it does`

`HumanResourceRemoveDaysOffTest`

* `removing a single day off takes it off the resource and notifies once`
* `removing a day off the resource does not have changes nothing and notifies nobody`
* `clearing the days off empties the list and notifies once, whatever the count`
* `clearing an empty list of days off notifies nobody`
* `the dialog's clear-all-and-rewrite still costs the same notifications` —
  walks every combination of M and N in `0..3`, pins the absolute event counts
  for `clearDaysOff()` so that an implementation firing once per interval would
  be caught, and pins from the other side that reaching into the handed-out list
  is rejected, removes nothing and notifies nobody.

`HumanResourceDaysOffViewTest`

* `the list handed out by getDaysOff cannot be modified` — `clear`, `add`,
  `removeAt` and `set` on the returned list all throw
  `UnsupportedOperationException`, and the resource is unchanged afterwards.
* `the list handed out by getDaysOff is a view and not a copy` — a day off added
  after the getter call shows up in the list fetched before it.

Without the notification the first test fails with `expected: <1> but was: <0>`,
and without the unmodifiable view the first test of `HumanResourceDaysOffViewTest`
fails with `Expected java.lang.UnsupportedOperationException to be thrown, but
nothing was thrown`. `./gradlew test --continue` is green on the branch: 370 tests
on master, 378 on the branch.

Two existing tests changed with the view: the regression test for the last
absence now removes through `clearDaysOff()` instead of `getDaysOff().clear()`,
and the M/N test turned its "old route" half into the assertion that the old
route is now rejected.

### DIFF
--- a/ganttproject-tester/test/net/sourceforge/ganttproject/resource/HumanResourceDaysOffTest.kt
+++ b/ganttproject-tester/test/net/sourceforge/ganttproject/resource/HumanResourceDaysOffTest.kt
@@ -1,0 +1,85 @@
+/*
+Copyright 2026 BarD Software s.r.o
+
+This file is part of GanttProject, an open-source project management tool.
+
+GanttProject is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+
+GanttProject is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with GanttProject.  If not, see <http://www.gnu.org/licenses/>.
+*/
+package net.sourceforge.ganttproject.resource
+
+import biz.ganttproject.core.time.CalendarFactory
+import biz.ganttproject.core.calendar.GanttDaysOff
+import biz.ganttproject.customproperty.CustomColumnsManager
+import net.sourceforge.ganttproject.roles.Role
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+import java.text.DateFormat
+import java.util.Date
+import java.util.Locale
+
+/**
+ * Adding and removing days off must be symmetric with respect to notification: whoever listens to
+ * a resource has to learn about both, because both change the resource's load distribution.
+ *
+ * Adding goes through [HumanResource.addDaysOff], which resets the loads and fires. Removing has no
+ * such entry point at all -- the only way is to mutate the list handed out by
+ * [HumanResource.getDaysOff], and that list is not observed.
+ */
+class HumanResourceDaysOffTest {
+  init {
+    // GanttDaysOff builds GanttCalendars, and those need a locale.
+    object : CalendarFactory() {
+      init {
+        setLocaleApi(object : CalendarFactory.LocaleApi {
+          override fun getLocale(): Locale = Locale.GERMANY
+          override fun getShortDateFormat(): DateFormat =
+            DateFormat.getDateInstance(DateFormat.SHORT, Locale.GERMANY)
+        })
+      }
+    }
+  }
+
+  private class CountingView : ResourceView {
+    var changed = 0
+    override fun resourceAdded(event: ResourceEvent) {}
+    override fun resourcesRemoved(event: ResourceEvent) {}
+    override fun resourceChanged(e: ResourceEvent) { changed++ }
+    override fun resourceAssignmentsChanged(e: ResourceEvent) {}
+    override fun resourceStructureChanged() {}
+    override fun resourceModelReset() {}
+  }
+
+  private fun september(day: Int): Date = CalendarFactory.createGanttCalendar(2026, 8, day).time
+
+  @Test
+  fun `removing the last day off notifies the listeners just as adding it does`() {
+    val manager = HumanResourceManager(null as Role?, CustomColumnsManager())
+    val person = manager.newHumanResource()
+    person.name = "Tester"
+    manager.add(person)
+
+    val view = CountingView()
+    manager.addView(view)
+
+    person.addDaysOff(GanttDaysOff(september(9), september(10)))
+    assertEquals(1, view.changed, "adding a day off must notify the listeners")
+
+    view.changed = 0
+    person.daysOff.clear()
+    assertEquals(
+      1, view.changed,
+      "removing the LAST day off must notify the listeners too -- otherwise nobody recalculates"
+    )
+  }
+}
--- a/ganttproject-tester/test/net/sourceforge/ganttproject/resource/HumanResourceDaysOffTest.kt
+++ b/ganttproject-tester/test/net/sourceforge/ganttproject/resource/HumanResourceDaysOffTest.kt
@@ -32,9 +32,10 @@ import java.util.Locale
  * Adding and removing days off must be symmetric with respect to notification: whoever listens to
  * a resource has to learn about both, because both change the resource's load distribution.
  *
- * Adding goes through [HumanResource.addDaysOff], which resets the loads and fires. Removing has no
- * such entry point at all -- the only way is to mutate the list handed out by
- * [HumanResource.getDaysOff], and that list is not observed.
+ * Adding goes through [HumanResource.addDaysOff], which resets the loads and fires. Removing the
+ * last one used to notify nobody: there was no entry point for it, and the only way was to mutate
+ * the list handed out by [HumanResource.getDaysOff], which nothing was watching. Removing now goes
+ * through [HumanResource.clearDaysOff], which fires just as adding does.
  */
 class HumanResourceDaysOffTest {
   init {
@@ -76,7 +77,7 @@ class HumanResourceDaysOffTest {
     assertEquals(1, view.changed, "adding a day off must notify the listeners")
 
     view.changed = 0
-    person.daysOff.clear()
+    person.clearDaysOff()
     assertEquals(
       1, view.changed,
       "removing the LAST day off must notify the listeners too -- otherwise nobody recalculates"
--- a/ganttproject/src/main/java/net/sourceforge/ganttproject/gui/GanttDialogPerson.java
+++ b/ganttproject/src/main/java/net/sourceforge/ganttproject/gui/GanttDialogPerson.java
@@ -165,7 +165,7 @@ public class GanttDialogPerson {
       return null;
     });
 
-    person.getDaysOff().clear();
+    person.clearDaysOff();
     for (DateInterval interval : myDaysOffModel.getIntervals()) {
       person.addDaysOff(new GanttDaysOff(interval.getStart(), interval.getEnd()));
     }
--- a/ganttproject/src/main/java/net/sourceforge/ganttproject/gui/GanttDialogPerson.java
+++ b/ganttproject/src/main/java/net/sourceforge/ganttproject/gui/GanttDialogPerson.java
@@ -36,6 +36,7 @@ import net.sourceforge.ganttproject.task.TaskManager;
 
 import javax.swing.*;
 import java.awt.event.ActionEvent;
+import java.util.List;
 
 public class GanttDialogPerson {
   private static final GanttLanguage language = GanttLanguage.getInstance();
@@ -191,8 +192,8 @@ public class GanttDialogPerson {
         super.remove(interval);
       }
     };
-    DefaultListModel<GanttDaysOff> daysOff = person.getDaysOff();
-    for (int i = 0; i < daysOff.getSize(); i++) {
+    List<GanttDaysOff> daysOff = person.getDaysOff();
+    for (int i = 0; i < daysOff.size(); i++) {
       GanttDaysOff next = daysOff.get(i);
       myDaysOffModel.add(DateInterval.Companion.createFromModelDates(next.getStart().getTime(),
           next.getFinish().getTime()));
--- a/ganttproject/src/main/java/net/sourceforge/ganttproject/io/VacationSaver.java
+++ b/ganttproject/src/main/java/net/sourceforge/ganttproject/io/VacationSaver.java
@@ -36,7 +36,7 @@ class VacationSaver extends SaverBase {
     for (HumanResource p : project.getHumanResourceManager().getResources()) {
       if (p.getDaysOff() != null)
         for (int j = 0; j < p.getDaysOff().size(); j++) {
-          GanttDaysOff gdo = (GanttDaysOff) p.getDaysOff().getElementAt(j);
+          GanttDaysOff gdo = p.getDaysOff().get(j);
           addAttribute("start", gdo.getStart().toXMLString(), attrs);
           addAttribute("end", gdo.getFinish().toXMLString(), attrs);
           addAttribute("resourceid", p.getId(), attrs);
--- a/ganttproject/src/main/java/net/sourceforge/ganttproject/resource/HumanResource.java
+++ b/ganttproject/src/main/java/net/sourceforge/ganttproject/resource/HumanResource.java
@@ -28,6 +28,8 @@ import net.sourceforge.ganttproject.task.ResourceAssignment;
 import net.sourceforge.ganttproject.task.Task;
 
 import javax.swing.*;
+import javax.swing.event.ListDataEvent;
+import javax.swing.event.ListDataListener;
 import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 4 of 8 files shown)


dbarashev:
@Natalie-the-technician maybe it is a good time to replace `DefaultListModel` with a standard `List` in the `HumanResource::getDaysOff()` return type and in the underlying field? `DefaultListModel` is a Swing class, and despite that it is a "model", it is actually a model of a UI control.

Natalie-the-technician:
Happy to. I looked at what depends on it first, and there is one consequence worth
naming before I change it, because it decides the shape of the result.

`DefaultListModel` is indeed never used as a UI model here. `GanttDialogPerson`
copies the intervals into its own `DateIntervalListEditor` model rather than
binding to it, and the four other readers only iterate:

| | uses |
|---|---|
| `GanttDialogPerson` (dialog setup) | `getSize()`, `get(i)` |
| `VacationSaver` | `size()`, `getElementAt(i)` |
| `LoadDistribution.processDaysOff` | `size()`, `get(i)` |
| `OverwritingMerger` | `size()`, `get(i)` |
| `HumanResource` copying constructor | `getSize()`, `get(i)` |

No test touches `getDaysOff()`. So the switch itself is small.

The consequence: this PR fixes the missing notification by observing the list —
a `ListDataListener` on the `DefaultListModel`. A plain `List` cannot be observed,
so the notification has to move into `addDaysOff` / `removeDaysOff` /
`clearDaysOff`, and then **the getter has to stop handing out something mutable**.
Otherwise the defect comes straight back: a caller mutating the returned list would
once again change the resource without resetting `myLoadDistribution` and without
firing, which is exactly what this PR is about.

That is a better end state than what is in the PR now, and it makes the three
methods the only way in and out. It does change a public signature, which is why I
did not go there unasked.

Two things I would rather have your preference on than guess:

1. `List<GanttDaysOff>` or `Collection<GanttDaysOff>` as the return type?
2. `Collections.unmodifiableList(new ArrayList<>(...))`, the way
   `HumanResourceManager.getResources()` does it at line 217 — or an unmodifiable
   view without the copy?

I will fold it into this PR rather than open a second one, since without the
unmodifiable getter the type change would leave the hole open.

dbarashev:
I believe it must be an unmodifiable view with the List type.

Natalie-the-technician:
Done — `getDaysOff()` now returns an unmodifiable `List` view.

The field behind it is a plain `ArrayList`, and `getDaysOff()` hands out
`Collections.unmodifiableList` over it, so callers keep seeing the resource but
cannot change it. The reading callers moved from `getSize()`/`getElementAt(i)` to
`size()`/`get(i)`; `LoadDistribution` and `ProjectFileExporter` were reading
through a raw `DefaultListModel` and lost their casts along the way.

Two places where this was more than a type swap:

1. **The notification moved into the methods.** A `DefaultListModel` could be
   watched with a `ListDataListener`, which is how the removal of the last day off
   came to notify its listeners; a plain `List` cannot be watched. `addDaysOff`,
   `removeDaysOff` and `clearDaysOff` now reset the load distribution and fire
   themselves. That is only sound because the getter turned tight at the same time
   — while the handed-out list was modifiable, a caller could change the resource
   without either happening, which is the defect this PR fixes.

2. **The event count is unchanged.** `DefaultListModel.clear()` does not fire on an
   empty list and `removeElement()` does not fire when nothing matched; `List` has
   no such behaviour, so `clearDaysOff()` returns early when the list is already
   empty and `removeDaysOff()` returns `false` without firing. The dialog's
   clear-all-and-rewrite therefore still costs `(M > 0 ? 1 : 0) + N` events, and
   the test walking every M and N in `0..3` still pins those absolute numbers.

Two new tests pin the return value itself: that modifying it throws
`UnsupportedOperationException`, and that it is a view rather than a copy (a day
off added after the call shows up in a list fetched before it). The first one
fails against the previous state of this branch with
`Expected java.lang.UnsupportedOperationException to be thrown, but nothing was
thrown`.

`./gradlew test --continue` is green: 376 tests before, 378 after, the two new
ones being the difference.

ganttproject/src/main/java/net/sourceforge/ganttproject/io/VacationSaver.java
```diff
@@ -36,7 +36,7 @@ void save(IGanttProject project, TransformerHandler handler) throws SAXException
     for (HumanResource p : project.getHumanResourceManager().getResources()) {
       if (p.getDaysOff() != null)
         for (int j = 0; j < p.getDaysOff().size(); j++) {
-          GanttDaysOff gdo = (GanttDaysOff) p.getDaysOff().getElementAt(j);
+          GanttDaysOff gdo = p.getDaysOff().get(j);
```
dbarashev, line 39: Can you please save days off into a local variable, to avoid calling it (and creating a read-only view) many times?

ganttproject-tester/test/net/sourceforge/ganttproject/resource/HumanResourceDaysOffTest.kt
dbarashev, line 1: Any sense to join these three test classes into a single one?

ganttproject/src/main/java/net/sourceforge/ganttproject/gui/GanttDialogPerson.java
```diff
@@ -165,7 +166,7 @@ private void applyChanges() {
       return null;
     });
 
-    person.getDaysOff().clear();
+    person.clearDaysOff();
```
dbarashev, line 169: maybe we can introduce setDaysOff() method which would replace the entire list (and send just one notification) ?

ganttproject/src/main/java/net/sourceforge/ganttproject/resource/LoadDistribution.java
```diff
@@ -57,10 +55,10 @@ public LoadDistribution(HumanResource resource) {
   }
 
   private void processDaysOff(HumanResource resource) {
-    DefaultListModel daysOff = resource.getDaysOff();
+    List<GanttDaysOff> daysOff = resource.getDaysOff();
     if (daysOff != null) {
```
dbarashev, line 59: This is redundant now

ganttproject/src/main/java/net/sourceforge/ganttproject/io/VacationSaver.java
```diff
@@ -36,7 +36,7 @@ void save(IGanttProject project, TransformerHandler handler) throws SAXException
     for (HumanResource p : project.getHumanResourceManager().getResources()) {
       if (p.getDaysOff() != null)
```
dbarashev, line 37: No more needed

ganttproject/src/main/java/net/sourceforge/ganttproject/resource/HumanResource.java
```diff
@@ -184,13 +202,58 @@ public Role getRole() {
...
+   * cannot be modified.
+   *
+   * The interval is matched the way the list matches it, that is by {@code Object.equals}. Note that
+   * GanttDaysOff only overloads {@code equals(GanttDaysOff)} and does not override
+   * {@code equals(Object)}, so an interval built afresh from the same two dates is NOT the one this
+   * resource holds. Pass an instance obtained from this resource.
+   *
+   * @return true if the interval was there and has been removed, false if there was nothing to do
+   */
+  public boolean removeDaysOff(GanttDaysOff gdo) {
```
dbarashev, line 221: This method is not yet used, right? Do we have any plans of using it?
Anyway, since we delegate removal to the list, it makes sense to define equals and hashCode methods in `GanttDaysOff`

dbarashev (commented):
Looks good. A few relatively minor comments below.

dbarashev:
@Natalie-the-technician review :arrow_up:

Natalie-the-technician:
Thanks for the review — the remaining points are addressed now.

**`HumanResource.java:221` — "since we delegate removal to the list, it makes sense to define `equals` and `hashCode` methods in `GanttDaysOff`"**

Done, and you were more right than it looked. `GanttDaysOff` did have an `equals`, but only as an overload — `public boolean equals(GanttDaysOff)`, never `equals(Object)`. `List.remove()` and `List.contains()` call `equals(Object)`, so they never reached it: removal by value silently did nothing, and an interval could only be removed by handing back the very object the resource had been given. The overload had no callers anywhere in the tree, so it is now the real override.

`hashCode()` is deliberately *not* built from the two `GanttCalendar`s' own hash codes. `GanttCalendar` overrides `equals(Object)` — comparing year, month and day only — without overriding `hashCode()`, so it inherits `GregorianCalendar.hashCode()`, which takes the time of day into account. Two equal `GanttCalendar`s can therefore have different hash codes, and a `GanttDaysOff` delegating to them would inherit that broken contract. Since a day off built from a `Date` keeps whatever time of day that `Date` carried, this is the ordinary case and not a contrived one. `hashCode()` is built from the year, month and day of both ends — exactly the fields `GanttCalendar.equals` looks at. There is a test for it, and it does fail if the implementation is changed to delegate.

**"This method is not yet used, right? Do we have any plans of using it?"**

Correct — `removeDaysOff()` has no caller in the tree. `git grep removeDaysOff` finds only its own definition, the javadoc references, and the tests. That has not changed with this round, and there is no concrete plan that needs it.

What it does have is a reason to exist that is independent of a caller: it is the counterpart of `addDaysOff()`. Before this PR, `getDaysOff()` handed out the live list and removal happened by mutating it, which notified nobody. Once the handed-out list became unmodifiable, every way of adding needed a matching way of removing, or the API would have been one-directional. `clearDaysOff()` covers "all of them" and is used by the properties dialog; `setDaysOff()` covers "replace them"; `removeDaysOff()` covers "this one", which is the granularity a resource-level undo or a levelling feature would want.

That said — if you would rather not carry an unused public method, say so and I will drop it. The tests for it would go with it, and `clearDaysOff()` plus `setDaysOff()` cover every caller that exists today.

**`GanttDialogPerson.java:169` — "maybe we can introduce `setDaysOff()` method which would replace the entire list (and send just one notification)?"**

Done. `HumanResource.setDaysOff()` replaces the whole list and notifies once, and `GanttDialogPerson.applyChanges()` uses it.

The measured effect, from the test that counts notifications on both ways side by side: replacing M intervals with N cost `(M > 0 ? 1 : 0) + N` notifications before — so 4 for the common three-in, three-out case — and costs 1 now. It is not just a smaller number: each of those notifications resets the resource's load distribution, so every partial state on the way, with some of the new intervals but not all of them, was announced to every listener as if it were real.

Two details worth flagging:

- It notifies once when the new intervals differ from the current ones and *not at all* when they do not, which is the rule `addDaysOff`, `removeDaysOff` and `clearDaysOff` already follow. Opening the dialog and pressing Ok without touching the days off is now silent. That comparison is by value and only works because of the `equals(Object)` above — the dialog rebuilds its intervals from the date pickers and hands over fresh objects.
- It copies the argument before clearing, because the argument may be the view `getDaysOff()` hands out, or a sublist of it. A sublist is the sharp case: it is a live window onto the resource's own list and is *not* equal to it, so the nothing-changed shortcut does not fire and the list really is cleared out from under the argument. Without the copy that throws `ConcurrentModificationException`; there is a test for it.
- It takes a `Collection` rather than a `List` so that it does not pair up with `getDaysOff()` into a single read-write property. The getter hands out an unmodifiable view of the resource's own list, which is not what the setter consumes, and Kotlin callers would otherwise see them as one `var`.

**`HumanResourceDaysOffTest.kt` — "Any sense to join these three test classes into a single one?"**

Yes. They are one class now. They were split by which method they pinned, but they share the whole fixture — the locale setup, the counting `ResourceView`, the manager-with-one-resource helper — and they describe one feature from several sides rather than three features.

No test was lost. Seven of the eight methods are carried over as they were; the eighth, `the dialog's clear-all-and-rewrite still costs the same notifications`, is absorbed into the new `setDaysOff` counting test, which runs the same M/N matrix and carries every one of its assertions. I checked that by diffing the assertion messages of the three original files against the merged one: 24 of the 25 are verbatim, and the 25th is a verbatim prefix of its replacement. The suite went from 279 tests to 285.

One test did change its meaning rather than its wording. `removing a day off the resource does not have changes nothing and notifies nobody` used to pass an interval with *the same dates* as one the resource held, and pinned the fact that it was still a different interval as far as the list was concerned. That was a consequence of the missing `equals(Object)`, so with this change it is no longer true. It now passes an interval the resource genuinely does not have, and a new test pins the opposite property: removal is by value, whoever built the interval.

**`LoadDistribution.java:59` — "This is redundant now" / `VacationSaver.java:37` — "No more needed"**

Both removed, along with the local-variable point at `VacationSaver.java:39`. `getDaysOff()` returns a final, never-null view, so there was nothing to guard against in either place. Both index loops are for-each loops now, which also takes `VacationSaver`'s three `getDaysOff()` calls per interval down to one.

Apologies for the delay on these two — they were marked as handled on my side but were still in the tree; thanks for the nudge.

---

The whole suite is green: 285 tests, 0 failures, 0 errors.

PR #2831, opened 2026-08-31
Ticking "Coordinator" in Task properties has no effect: the click never reaches the assignment

Natalie-the-technician:
## What is wrong

The **Coordinator** column in *Task properties → Resources* can be ticked, the
tick appears, and the dialog closes without complaint — but the value never
arrives at the `ResourceAssignment`. The project is saved with
`responsible="false"`, and reopening the tab shows the box empty again.

Nothing is reported. The only way to notice is to look into the saved file or to
reopen the tab.

## How to reproduce

1. Open a project, assign two resources to a task.
2. Task properties → *Resources*. Tick **Coordinator** for the second resource.
3. Press OK, save the project.
4. Look into the `.gan` file: the `<allocation>` element still says
   `responsible="false"`.
5. Reopen the tab: the box is empty.

## Why this is the cause

`CheckBoxTableCell.forTableColumn(column)` does not start an edit at all. It
binds the check box **bidirectionally** to whatever the cell value factory
returned, so the `setOnEditCommit` handler installed next to it is never fired
and is dead code.

Read off the bytecode of `javafx-controls` 21 rather than inferred:

* the class contains no `startEdit`, no `commitEdit` and no `cancelEdit`;
* `forTableColumn(TableColumn)` compiles to `forTableColumn(null, null)`, so no
  `selectedStateCallback` is installed;
* `getSelectedProperty()` therefore falls back to
  `getTableColumn().getCellObservableValue(index)`;
* `updateItem` binds the check box to it with
  `BooleanProperty.bindBidirectional`.

The second half of the defect is the factory:

```kotlin
setCellValueFactory { SimpleBooleanProperty(it.value.assignment?.isCoordinator ?: false) }
```

It hands out a **fresh** `SimpleBooleanProperty` on every call. The bidirectional
binding therefore writes the click into an object that nobody reads again, and
the value is discarded when the cell is next updated.

Either half alone would be harmless: a factory returning a stable property would
work with the binding, and a real edit would reach `setOnEditCommit`. It is the
combination that swallows the click silently.

## What this changes

The factory now returns a property that carries a listener writing the new value
into the assignment — the same write the table model performs for this column.
The dead `setOnEditCommit` block is removed, so there is exactly one write path
and not two.

## Tests

`TaskResourcesPanelTest`

* `ticking coordinator reaches the assignment`
* `unticking coordinator reaches the assignment as well`
* `a newly added assignment takes the tick too`

Both directions are covered, and the third case works because the write reaches
the mutator's stub, whose value is copied over in
`ResourceAssignmentCollectionImpl.commit`.

Without the fix the first test fails with `expected: <true> but was: <false>`.
`./gradlew test --continue` is green on the branch; the three tests above are the
only ones added and no existing test changes.

### DIFF
--- a/ganttproject/src/main/java/net/sourceforge/ganttproject/gui/taskproperties/TaskResourcesPanel.kt
+++ b/ganttproject/src/main/java/net/sourceforge/ganttproject/gui/taskproperties/TaskResourcesPanel.kt
@@ -170,14 +170,12 @@ class TaskResourcesPanel(
         prefWidth = 80.0
       }
 
-      // Coordinator column
+      // Coordinator column. The check box is bound to the value which the cell value factory
+      // returns, so that value has to be the one which writes into the assignment. See
+      // coordinatorProperty below.
       val coordinatorCol = TableColumn2<ResourceAssignmentRow, Boolean>(i18n.formatText("coordinator")).apply {
-        setCellValueFactory { SimpleBooleanProperty(it.value.assignment?.isCoordinator ?: false) }
+        setCellValueFactory { coordinatorProperty(it.value.assignment) }
         cellFactory = CheckBoxTableCell.forTableColumn(this)
-        setOnEditCommit { event ->
-          model.setValueAt(event.newValue, event.tablePosition.row, 3)
-          model.refreshTable()
-        }
         isEditable = true
         prefWidth = 100.0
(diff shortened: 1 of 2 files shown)


dbarashev:
Thanks for the fix!

PR #2832, opened 2026-08-31
Write recent colours before <filters> so that no view option is lost

Natalie-the-technician:
### What happens

`parseXmlProject` keeps only the **last uninterrupted run** of `<option>` elements
inside a `<view>`. As soon as an element of a different kind — `<filters>` or
`<timeline>` — stands between two runs, everything before it is dropped without a
word. The file on disk still holds every option; the program no longer sees them.

`ViewSaver.save` writes `color.recent` after `writeFilters(...)`. So in any project
that has recent colours the last run consists of that single option, and the up to
nine written before it are gone on the next load: four `taskLabel*`, four
`filter.*`, and `divider`, the position of the splitter between the task table and
the chart. What a user sees is a splitter that keeps jumping back and label
settings that do not stick, with nothing to connect that to recent colours.

The same loss can be seen on the sample project that ships with GanttProject. In
`ganttproject-builder/HouseBuildingSample.gan` the `gantt-chart` view carries six
`<option>` elements with a `<timeline>` standing between them, and exactly one is
read back:

```
<option> elements in the view:   6
view id=gantt-chart read back:   1  ->  [color.recent]
```

(That file was written by an older version, so its element order is not the one the
current saver produces — but it is opened by the current reader, and it loses five
of its six options.)

### The change

`writeRecentColors(handler)` moves before `writeFilters(handler, ...)`. All options
of the view then form a single run and `<filters>` is the last element.

### Why the writing side and not the reader

* It is one line, and no reader has to change.
* Older versions read the new order at least as well as the old one. Since the
  Jackson reader (7767f82fb, 2022-04-01) a single run is exactly the case they
  handle best; before that, `OptionTagHandler` matched every `<option>` by its `id`
  regardless of position, so for those versions the two orders are equivalent.
  Reading the same two documents with the reader built at 08564e556:

```
old order (options, <filters>, color.recent): 1 -> [color.recent]
new order (options, color.recent, <filters>): 5 -> [taskLabelUp, taskLabelDown, divider, filter.completedTasks, color.recent]
```

* Files written by older versions keep losing what they lose: this fixes writing,
  not reading. Whether the reader should also be made tolerant is a separate
  question and is not proposed here.

### Checks

`ViewOptionOrderTest` runs the real `ViewSaver` and reads the document back with
the real reader. Without the change, `in the order the saver writes today no option
is lost` fails with

```
expected: <[taskLabelUp, taskLabelDown, divider, filter.completedTasks, color.recent]>
 but was: <[color.recent]>
```

The second check, `the reader drops every option that stands before another
element`, pins the reader behaviour the order rests on, so the reason for it is not
lost if anyone puts the old order back.

`./gradlew test`: 370 tests before, 372 after, all green.

### DIFF
--- a/ganttproject/src/main/java/net/sourceforge/ganttproject/io/ViewSaver.java
+++ b/ganttproject/src/main/java/net/sourceforge/ganttproject/io/ViewSaver.java
@@ -54,8 +54,13 @@ class ViewSaver extends SaverBase {
     writeTimelineTasks(facade, handler);
     new OptionSaver().saveOptionList(handler, facade.getGanttChart().getTaskLabelOptions().getOptions());
     new OptionSaver().saveOptionList(handler, ganttViewProvider.getOptions());
-    writeFilters(handler, taskFilterManager);
+    // The recent colours are written here, before <filters>, so that all <option> elements of this
+    // view form a single uninterrupted run. The reader keeps only the last such run and drops
+    // whatever stands before an element of a different kind without a word. With color.recent
+    // written after <filters> that run consisted of color.recent alone, and the task label options,
+    // the view options and the divider position were lost on the next load.
     writeRecentColors(handler);
+    writeFilters(handler, taskFilterManager);
     endElement("view", handler);
 
     addAttribute("id", "resource-table", attrs);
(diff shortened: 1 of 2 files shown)


dbarashev:
Thanks for the fix!

PR #2837, opened 2026-09-20
Removing a baseline leaves its temporary file behind on Windows

Natalie-the-technician:
### The problem

Every baseline is kept in a temporary file. Removing the baseline is supposed to remove the file. On Windows it does not: the file stays until the JVM exits, and nothing says so.

### Cause

Two lines, one at each end.

`GanttPreviousState.BaselineSaver.save()` hands the `File` to `StreamResult`:

```java
StreamResult result = new StreamResult(file);
```

The JDK serialiser then opens a stream of its own behind that `File` and never closes it.

`GanttPreviousState.remove()` throws away the answer:

```java
public void remove() {
  myFile.delete();
}
```

Windows refuses to delete a file that is still open, so `delete()` returns `false` — and the return value goes nowhere. On Linux the delete succeeds whether a handle is open or not, which is why this was never noticed there.

`deleteOnExit()` is called when the file is created, so the files do go — but only at the end of the session, and only if the handle has been released by then.

### The fix

`save()` opens the stream and closes it:

```java
try (OutputStream output = new FileOutputStream(file)) {
  TransformerHandler handler = createHandler(new StreamResult(output));
  ...
}
```

and `remove()` answers instead of shrugging:

```java
public boolean remove() {
  if (myFile == null || !myFile.exists()) {
    return true;
  }
  if (myFile.delete()) {
    return true;
  }
  LOG.error("Failed to delete the temporary file of baseline {}: {}", ...);
  return false;
}
```

`void` to `boolean` is source-compatible for the two existing callers (`BaselineDialogAction` and `GanttProject`), neither of which looks at the result today. The failure goes to the logger, not to `System.err`.

### The test

`BaselineTempFileTest` drives `GanttPreviousState` directly. It cannot redirect `java.io.tmpdir` — `File.createTempFile` reads that property once when `java.io.File` is loaded — so it notes which `_GanttProject_ps_*.gan` files exist before the baseline is created and works with the difference. No reflection, no access to private state.

**Seen red before it was seen green.** Against unchanged code on Windows:

```
BaselineTempFileTest > the temporary file of a removed baseline is gone  FAILED
  expected: <[]> but was: <[C:\Users\...\Temp\_GanttProject_ps_27266781054304562476573.gan]>
```

Five checks, and each one earns its place. Four separate mutilations of the fix were applied one at a time and measured:

| Mutilation | caught by |
|---|---|
| hand the `File` to `StreamResult` again, leaving the stream open | *the temporary file of a removed baseline is gone*, *removing a baseline reports success* |
| call `delete()` but discard the result and return `true` | *a failed delete is reported, not swallowed*, *a failed delete names the file in the log* |
| never call `delete()` at all | *the temporary file of a removed baseline is gone* and both failure-path checks |
| return the right answer but log nothing | *a failed delete names the file in the log* |

The two failure-path checks turn the baseline's file into a directory that is not empty. `File.delete()` refuses that on every platform, so those two run on Linux as well — measured, not argued. On Linux all five pass, and with `remove()` mutilated to discard the result and return `true`, exactly those two turn red:

```
BaselineTempFileTest > a failed delete names the file in the log()   FAILED
BaselineTempFileTest > a failed delete is reported, not swallowed()  FAILED
5 tests completed, 2 failed
```

The leak check itself is Windows-only in effect: on Linux it passes with or without the fix, because the platform does not care about the open handle. That is the whole shape of the bug, so I do not think it should be marked as skipped elsewhere — it simply cannot fail where the platform is forgiving.

`:ganttproject:test` as a whole: 91 checks, 0 skipped, 0 red — the same number on Windows and on Linux.


dbarashev:
An alternative take on the same bug: #2838 removes the temporary-file mechanism entirely instead of fixing the leak.

The observation behind it: `GanttPreviousState.myTasks` already holds the full snapshot for the object's whole lifetime, and the temp file was only ever a round trip back to that same list — write XML out, parse the same XML back. Since the snapshot is immutable (`GanttPreviousStateTask` holds plain values, `GanttCalendar` has no public mutators), keeping it in memory is safe, and every deleted-file bug on Windows ceases to exist along with the files.

As a bonus, it fixes a latent crash: `XmlProjectImporter` (used by the Colloboque server) already creates baselines with no file at all, and `load()` on one of those would throw today.

Happy to see either approach land — this one fixes the reported bug minimally and safely; the other deletes ~100 lines and the whole class of issues.

Natalie-the-technician:
Switching to the approach in #2838

PR #2838, opened 2026-09-21
Keep baselines in memory instead of temporary files

dbarashev:
### The problem

Every baseline is kept in a temporary file. On Windows the file cannot be deleted while the JVM lives, because the stream that wrote it was never closed — removing a baseline or closing a project leaves `_GanttProject_ps_*.gan` files behind in the system temp directory.

This is an alternative to #2837: instead of fixing the leak, it removes the mechanism that leaks.

### Why the file existed at all

It didn't need to. `GanttPreviousState.myTasks` already holds the complete snapshot for the object's whole lifetime. The file was written once from that list and never changed; every `load()` re-parsed the file and returned exactly what `myTasks` already contained. A pure round trip — serialize to disk, parse the same bytes back.

The snapshot is immutable: `GanttPreviousStateTask` holds plain values, and `GanttCalendar` has no public mutators, so sharing the in-memory list is safe.

### The change

`GanttPreviousState` keeps its tasks and answers `getTasks()` with them. Gone:

- `init()` / `saveFile()` / `remove()` / `load()` and the `_GanttProject_ps_*` files
- the `deleteOnExit()` backstop and the cleanup loops in `BaselineDialogAction` and `GanttProject.close()`
- `PreviousStateTasksTagHandler`, whose only live use was parsing those files (its project-parsing constructor was already dead code)
- the `IOException` path in the baseline dialog: creating a baseline can no longer fail

Callers now use the in-memory list directly: the chart (`GanttGraphicArea.setBaseline`), project save (`HistorySaver`), and project load (`BaselineSerializer`).

A latent crash goes away as a side effect: `XmlProjectImporter` (used by the Colloboque server) already created baselines with no file at all — any `load()` on such a baseline would have thrown.

### Compatibility

The project file format is unchanged. Baselines are still embedded in `.gan` on save by `HistorySaver` and read back on open by `BaselineSerializer`; only the session-scratch temp files disappear.

### The test

`BaselineRoundTripTest` drives the real `HistorySaver` and the real `BaselineSerializer` back to back:

- *a baseline survives a save and load round trip* — name, id, start, duration, milestone and summary flags of three tasks come back unchanged
- *loading a baseline touches no temporary file* — snapshots the temp dir before and after a load and asserts no `_GanttProject_ps_*` appears, so the in-memory invariant is enforced, not just documented

`:ganttproject:test`: 88 checks, 0 skipped, 0 failures. Full `assemble` builds.


Natalie-the-technician:
Your approach is the better one. I measured both branches against `master` (`2085c2d2f`);
the numbers are below.

**The crash is real, and it is worse than you described.** Reproduced on today's `master`:
`XmlProjectImporter` never calls `init()`, so `myFile` stays null and every path that reads
such a baseline ends in

```
java.lang.IllegalArgumentException: File cannot be null
    at GanttPreviousState.load(GanttPreviousState.java:105)
    at HistorySaver.saveBaseline(HistorySaver.java:45)
```

It is not only `load()`. `GanttXMLSaver` calls `HistorySaver.save` (line 137), and
`ProjectFileUpdater.updateProjectFile` writes the project back out that way — so a project
with a single baseline cannot round-trip through Colloboque today. #2837 does not fix this:
it guards `myFile == null` in `remove()` only and leaves `load()` untouched.

**Keeping baselines in memory costs 493 bytes per entry.** Measured at n = 25k/100k/400k
after a settled heap, spread 0.3%; 469 bytes before an entry is first read, 493 once the
`GregorianCalendar` has computed its fields. For comparison, a live task in the same heap
costs about 1.8 kB (marginal, from n = 1000 vs n = 4000), and one entry in the temporary
file costs 92 bytes today. With the 15-per-kind cap our fork applies, that is ~14.8 kB per
task across 30 baselines. Against the `-Xmx2048m` GanttProject ships with, baselines alone
reach a quarter of the heap at roughly 35 000 tasks. That is far outside normal use, so
memory is not an argument against this.

**One thing in the PR description does not hold:** `GanttCalendar` has 17 public instance
mutators, all inherited from `GregorianCalendar` and none overridden —
`set(int,int)`, `setTime`, `setTimeInMillis`, `add`, `roll`, `clear`, `setWeekDate` and the
rest. Your second commit clearly knows this: the `clone()` in `createTasks` and the comment
in `BaselineSnapshotTest` say exactly that, so the branch is ahead of its own description.

**What the clone does not cover is the other direction.** It stops the baseline from
following its task. It does not stop a consumer of the baseline from reaching in, because
`getTasks()` hands out `myTasks` itself and each entry hands out its `GanttCalendar` itself.
I ran the same assertions, word for word, in both worktrees — `load()` on `master`,
`getTasks()` on your branch, nothing else different:

| | master | #2838 |
|---|---|---|
| the baseline keeps its date when a caller changes what it handed out | pass | **fail** — 2026-09-11 becomes 2027-09-11 |
| the baseline keeps its tasks when a caller empties the list it handed out | pass | **fail** — 1 becomes 0 |
| the saved project keeps the date afterwards | pass | **fail** — `start="2027-09-11"` is written |

An `add(Calendar.YEAR, 1)` on what `getTasks()` returns moves the stored baseline
permanently and ends up in the project file. On `master` the same call is harmless, because
`load()` re-parses: the temporary file happens to double as a defensive copy, which is the
one property that goes away with the files.

No consumer in the tree does this today — `renderBaseline` only calls `getStart().getTime()`,
and `Calendar.getTime()` returns a fresh `Date` — so it is a trap rather than a live bug.
Two lines close it:

```java
public List<GanttPreviousStateTask> getTasks() {
  return Collections.unmodifiableList(myTasks);
}
```
```java
public GanttCalendar getStart() {
  return myStart.clone();
}
```

Both assertions above turn green with that, and nothing in the current tree notices: the
renderer calls `getStart()` once per matching row.

**On the tests in #2837:** they do not fail on your branch, they do not compile — 12 errors,
`init`, `saveFile` and `remove` all unresolved. All five test behaviour that stops existing
along with the files, and the one that checked the written file's content has a better
successor in your `BaselineRoundTripTest`, which goes through the real `HistorySaver` and
the real `BaselineSerializer` instead of grepping the file text. The file-counting trick in
`loading a baseline touches no temporary file` is the same idea turned around, which is the
right place for it.

So: #2838 should land. The three assertions above are the only thing from #2837 worth
carrying over — the rest of it is obsolete either way.

